PR #1, opened 2019-04-10
Write initial Manager and Package classes

njgheorghita:
## What was wrong?

Some things to consider..

- Layout of `ethpm.lock`
- Is there a better name for `Package` class (so it doesn't get confused with `ethpm.Package`

#### Cute Animal Picture
image


njgheorghita:
@pipermerriam Having some trouble getting ipfs up and running in integration tests - but if you have some time to give this a :eyes: that'd be great!

pipermerriam:
Hrm, can we compromise in CI for now and use the IPFS gateway or a mock-based solution?

__main__.py
```diff
@@ -0,0 +1,26 @@
...
+
+__version__ = pkg_resources.require("ethpm-cli")[0].version
+print(f"EthPM CLI v{__version__}\n")
+
+parser = argparse.ArgumentParser(description='ethpm-cli')
+subparsers = parser.add_subparsers(help='commands', dest='command')
+
+install_parser = subparsers.add_parser('install', help='Install uri')
+install_parser.add_argument('uri', action='store', type=str, help='IPFS / Github / Registry URI of package you want to install')
+install_parser.add_argument('-pkgs_dir', action='store', type=str, help='Path to specific ethpm_packages dir')
```
pipermerriam: This is likely something worth updating the tactical manual with a standard convention.

In the past, I've tended to use the following convention which seems common.

- Long flags are `--packages-dir`  (note double dash prefix)
- Shortcut aliases are `-p` (note single dash and single letter)

Underscores seem out of place here and single dash prefixes on non-single letter flags always seem weird to me but I'm willing to be convinced otherwise as it allows other parts of the code to print information out to the terminal without litering them with print statements as well.

__main__.py
```diff
@@ -0,0 +1,26 @@
+from docopt import docopt
+import argparse
+import os
+import pkg_resources
+
+from ethpm_cli.manager import Manager
+
+__version__ = pkg_resources.require("ethpm-cli")[0].version
+print(f"EthPM CLI v{__version__}\n")
```
pipermerriam: In trinity we use the facilities from the `logging` module to do this, hooking the `StreamHandler` up to `sys.stdout` and `sys.stderr`.  I think this is a better approach than `print` statements.

ethpm_cli/manager.py
```diff
@@ -0,0 +1,87 @@
...
+
+from eth_utils.toolz import assoc
+from ethpm.typing import URI
+
+from ethpm_cli.exceptions import InstallError
+from ethpm_cli.package import Package
+from ethpm_cli._utils.ipfs import get_ipfs_backend
+
+
+class Manager:
```
pipermerriam: This is an anti-pattern to watch out for.  A class with only an `__init__` and a single method.  These are just functions disguised as classes.  This could and probably should be converted to a single function.

ethpm_cli/manager.py
```diff
@@ -0,0 +1,87 @@
...
+    manifest: Manifest, parent_dir: Path, ipfs: BaseIPFSBackend
+) -> None:
+    (parent_dir / "src").mkdir()
+    for path, source in manifest["sources"].items():
+        if is_ipfs_uri(source):
+            source_contents = to_text(ipfs.fetch_uri_contents(source)).rstrip("\n")
+        else:
+            # for inlined sources
+            source_contents = source
+        target_file = parent_dir / "src" / path
```
pipermerriam: There should be validation to ensure that this cannot result in writing to a directory that is not strictly *under* `parent_dir / "src"`.  We should have tests for this to be extra extra sure this cannot happen.

ethpm_cli/manager.py
```diff
@@ -0,0 +1,87 @@
...
+        ethpm_lock.touch()
+    new_pkg_data = pkg.generate_ethpm_lock()
+    new_lock = assoc(old_lock, pkg.alias, new_pkg_data)
+    ethpm_lock.write_text(json.dumps(new_lock, sort_keys=True, indent=4))
+
+
+def write_build_deps_to_disk(build_deps: Dict[str, str], ethpm_dir: Path) -> None:
+    dep_manager = Manager(ethpm_dir)
+    for dep in build_deps.values():
+        dep_manager.install(dep)
```
pipermerriam: I can see a bit how this class could end up being useful but I'd argue YAGNI for now until it has more APIs or useful properties that you are making use of.

ethpm_cli/manager.py
```diff
@@ -0,0 +1,87 @@
...
+        self.ethpm_dir = process_target_dir(target_dir)
+        self.ethpm_lock = self.ethpm_dir / "ethpm.lock"
+        self.ipfs_backend = get_ipfs_backend(ipfs)
+
+    def install(self, target_uri: URI, alias: str = None) -> None:
+        target_pkg = Package(target_uri, alias, self.ipfs_backend)
+        pkg_path = self.ethpm_dir / target_pkg.alias
+        if pkg_path.is_dir():
+            raise InstallError(f"{target_pkg.alias} is already installed.")
+        pkg_path.mkdir()
```
pipermerriam: There should be validation here to ensure that `pkg_path` is strictly *under* `self.ethpm_dir` to prevent the ability for something to write data outside of where it should.

ethpm_cli/package.py
```diff
@@ -0,0 +1,120 @@
...
+        yield "target_uri", self.target_uri
+        yield "registry_address", self.registry_address
+        yield "alias", self.alias
+        yield "resolved_version", self.manifest["version"]
+        yield "resolved_package_name", self.manifest["package_name"]
+
+
+def determine_alias(pkg_name: str, alias: str = None) -> str:
+    if alias:
+        validate_package_name(alias)
```
pipermerriam: I think this validation is happening at the wrong level.  At this stage it should probably be assumed to be valid with validation having occurred higher up (like in the `Config` object, or as soon as the CLI args are parsed and you know what the alias is).

tests/core/test_import.py
```diff
@@ -1,4 +0,0 @@
-
-
-def test_import():
-    import <MODULE_NAME>  # noqa: F401
```
pipermerriam, line 4: This is actually a nice smoke test to keep around to detect really high level errors (that would admittedly manifest in other tests too)

pipermerriam (commented):
One thing you might consider is a `Config` object that holds all the stateful stuff such as IPFS backends, things the user declared from the command line, etc and passing this object around and down through the calls where appropriate (typically only appropriate in the high level APIs).

Another thing that seems to be missing is installing to a temporary directory and only updating the local dir once everything has completed successfully.  This keeps you from accidentally leaving things in an inconsistent state.  One thing you get from this is simplified code since you never have to check for whether a directory or file already exists since you are always working with a completely clean set of directories.

ethpm_cli/manager.py
njgheorghita: Assuming that `alias` is a valid package name (which is provided from cli arg validation) - Is it not safe to assume that `pkg_path` is always strictly under `self.ethpm_dir` without doing additional validation? Or am I missing something?

ethpm_cli/manager.py
njgheorghita: Gotcha - I ended up moving some functions to this class - since it was easier to handle the ipfs backend by setting it as an attribute to the `Manager` class. What are your thoughts on setting the ipfs backend in an environment variable? It might be simpler than passing it back/forth between the `Manager` and `Package` classes? And could simplify some of the code

njgheorghita: Ahh nvm this - just saw the `Config` comment

pipermerriam: Sounds like you found a way forward on this, however, I am almost always :+1: on environment variable based configuration.

ethpm_cli/manager.py
pipermerriam: It's more that the validation of `alias` is implicit and since this would be a major-ish security vulnerability I'd like to have this explicitely validated independently.

njgheorghita:
@pipermerriam 
> Another thing that seems to be missing is installing to a temporary directory and only updating the local dir once everything has completed successfully.

I'm having some troubles figuring out how to implement this. It makes sense that we want to use a tmpdir to keep things neat. My problems seem to be around `ethpm.lock`. It doesn't seem to be as simple as write all the package assets to a tmpdir and then copy over. The parent `ethpm.lock` and all the subsequent in build dependencies are stateful themselves (i.e. needs updating with every additional pkg that's installed on its level).

Say i'm installing a package with 3+ levels of build dependencies. Is the idea that each level will have it's own tmpdir - which gets resolved into its parent's tmpdir - and finally copied over to the real `ethpm_dir` once the tree is resolved?

Another complication comes with installing a package into an existing `ethpm_dir`. Would we copy the existing `ethpm_dir` over to the `tmp_dir` before installing, and then completely replace the existing `ethpm_dir` upon completion?

Sorry for the word dump, not sure if my problems make sense, or if i'm missing something.

njgheorghita:
@pipermerriam ping for another :eyes:

__main__.py
```diff
@@ -0,0 +1,59 @@
...
+from pathlib import Path
+from ethpm_cli.manager import Manager
+from ethpm_cli.validation import validate_cli_args, parse_cli_args
+
+
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+handler = logging.StreamHandler(sys.stdout)
+handler.setLevel(logging.INFO)
+logger.addHandler(handler)
```
pipermerriam: This is probably best relegated to a small function and then called from within whatever the `main` executable is that gets called after arguments are parsed.  See below.

__main__.py
```diff
@@ -0,0 +1,59 @@
...
+)
+args = parser.parse_args()
+
+
+command = args.command
+if command == "install":
+    validate_cli_args(args)
+    config = Config(args)
+    package = Package(args.target_uri, args.alias, config.ipfs_backend)
+    install_package(package, config)
```
pipermerriam: Noting that this has no `else` clause and will silently bypass this if some *other* command is called....  That doesn't seem ideal.  maybe there should be an `else` that raises an exception?

__main__.py
```diff
@@ -0,0 +1,59 @@
...
+    dest="local_ipfs",
+    action="store_true",
+    help="Flag to use locally running IPFS node.",
+)
+args = parser.parse_args()
+
+
+command = args.command
+if command == "install":
+    validate_cli_args(args)
```
pipermerriam, line 57: I would recommend gathering the body of this `if` block and wrapping it in some sort of `main` function that you call with the parsed `args`.

ethpm_cli/_utils/testing.py
```diff
@@ -0,0 +1,35 @@
+import filecmp
+import os.path
+
+
+def are_dir_trees_equal(dir1: str, dir2: str) -> bool:
```
pipermerriam: I think a slighty more comforming name to our existing conventions would be `check_dir_trees_equal`.

ethpm_cli/_utils/testing.py
```diff
@@ -0,0 +1,35 @@
+import filecmp
```
pipermerriam, line 1: I love learning about stdlib tools that I was not previously aware of.

ethpm_cli/install.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    Class to manager CLI config options
+    - IPFS Backend
+    - Target ethpm_dir
+    """
+
+    def __init__(self, args: Namespace) -> None:
+        self.ipfs_backend = get_ipfs_backend(args.local_ipfs)
+        if args.ethpm_dir is None:
+            self.ethpm_dir = Path.cwd() / "ethpm_packages"
```
pipermerriam: The string `"ethpm_packages"` should probably be defined somewhere as a constant.

ethpm_cli/install.py
```diff
@@ -0,0 +1,104 @@
...
+            self.ethpm_dir.mkdir()
+        else:
+            self.ethpm_dir = args.ethpm_dir
+
+
+def install_package(pkg: Package, config: Config) -> None:
+    if (config.ethpm_dir / pkg.alias).is_dir():
+        raise InstallError(f"{pkg.alias} is already installed.")
```
pipermerriam: Maybe a more accurate message would be something like:

> Installation conflict: A directory already exists at the install location for the package "{pkg.name}" aliased to "{pkg.alias}" on the filesystem at `{config.ethp_dir / pkg.alias}`.

ethpm_cli/install.py
```diff
@@ -0,0 +1,104 @@
...
+        self.ipfs_backend = get_ipfs_backend(args.local_ipfs)
+        if args.ethpm_dir is None:
+            self.ethpm_dir = Path.cwd() / "ethpm_packages"
+            self.ethpm_dir.mkdir()
+        else:
+            self.ethpm_dir = args.ethpm_dir
+
+
+def install_package(pkg: Package, config: Config) -> None:
+    if (config.ethpm_dir / pkg.alias).is_dir():
```
pipermerriam: Would it be more appropriate to just do an `exists` check here? since any name conflict is likely an error case?

ethpm_cli/install.py
```diff
@@ -0,0 +1,104 @@
...
+    install_tmp_package(pkg, tmp_pkg_dir, config.ipfs_backend)
+
+    # Copy temp pacakge directory to ethpm dir namespace
+    dest_pkg_dir = config.ethpm_dir / pkg.alias
+    validate_parent_directory(config.ethpm_dir, dest_pkg_dir)
+    shutil.copytree(tmp_pkg_dir, dest_pkg_dir)
+    update_ethpm_lock(pkg, (config.ethpm_dir / "ethpm.lock"))
+
+
+def install_tmp_package(
```
pipermerriam: Maybe instead of `install_tmp_package` this utility could be more accurately named `write_package_installation_files` since the temporary directory piece is just an implementation detail that this function actually doesn't need to know anything about.

ethpm_cli/install.py
```diff
@@ -0,0 +1,104 @@
...
+    tmp_ethpm_lock = tmp_pkg_dir.parent / "ethpm.lock"
+    update_ethpm_lock(pkg, tmp_ethpm_lock)
+
+
+def write_sources_to_disk(
+    pkg: Package, pkg_dir: Path, ipfs_backend: BaseIPFSBackend
+) -> None:
+    for path, source in pkg.manifest["sources"].items():
+        if is_ipfs_uri(source):
+            source_contents = to_text(ipfs_backend.fetch_uri_contents(source)).rstrip(
```
pipermerriam: I think that resolution of the sources should happen before we attempt to write any of them.  i.e. we should fully resolve the full tree of source files and then this function gets a mapping of `Dict[Path, str]` where the `str` part is known to be the content.

ethpm_cli/install.py
```diff
@@ -0,0 +1,104 @@
...
+    write_build_deps_to_disk(pkg, tmp_pkg_dir, ipfs_backend)
+    tmp_ethpm_lock = tmp_pkg_dir.parent / "ethpm.lock"
+    update_ethpm_lock(pkg, tmp_ethpm_lock)
+
+
+def write_sources_to_disk(
+    pkg: Package, pkg_dir: Path, ipfs_backend: BaseIPFSBackend
+) -> None:
+    for path, source in pkg.manifest["sources"].items():
+        if is_ipfs_uri(source):
```
pipermerriam: I just realized it is impossible to directly embed a source file who's contents are a URI since the spec requires instrospection of the actual text to determine if it is a content URI or the actual source and you cant distinguish between the two if the file contents are a content URI.....

ethpm_cli/install.py
```diff
@@ -0,0 +1,104 @@
...
+        validate_parent_directory((pkg_dir / "src"), target_file)
+        target_file.write_text(source_contents)
+
+
+def write_build_deps_to_disk(
+    pkg: Package, pkg_dir: Path, ipfs_backend: BaseIPFSBackend
+) -> None:
+    if "build_dependencies" in pkg.manifest:
+        child_ethpm_dir = pkg_dir / "ethpm_packages"
+        if not child_ethpm_dir.is_dir():
```
pipermerriam: I think the next line will fail loudly if there happens to be a file in this location but that file is not a directory.  Is this intentional?

pipermerriam (commented):
Impartial review.  Will finish when I get into the office.

__main__.py
njgheorghita: argparse will raise an exception if any unsupported command is called, but I can still add an else just in case

ethpm_cli/install.py
njgheorghita: If I understand right, when we're writing to the tmpdir we don't have to really worry about that happening? In fact, we can remove the check on line 86 since if a manifest is valid, there will never be build dependencies with conflicting names.

ethpm_cli/install.py
njgheorghita: Right, I'm not sure I get the consequences of this though - is there a use case where somebody might want to write the a source file that contains a uri to disk rather than the resolved contents?

__main__.py
pipermerriam: In cases like that where we *implicitly* know something can't happen but it may not be 100% clear when reading the code I've gotten in the habit of putting a `raise Exception("INVARIANT")` to be explicit that the code path should be unconditionally unreachable.

ethpm_cli/install.py
pipermerriam: No, probably benign.  I just dislike it because it forbids something that *should* be possible despite it not having any clear use case.

ethpm_cli/install.py
pipermerriam: your call.

ethpm_cli/validation.py
```diff
@@ -0,0 +1,51 @@
...
+        and not is_valid_content_addressed_github_uri(uri)  # noqa: W503
+    ):
+        raise UriNotSupportedError(
+            f"Target uri: {uri} not a currently supported uri. "
+            "Target uris must be one of: ipfs, github blob, or registry."
+        )
+
+
+def validate_alias(alias: str) -> None:
+    if alias:
```
pipermerriam: IIUC this `if` statement is to allow for a `None` value?

Two things.

1. This would allow the empty string.  Is that intentional?  Also other falsy values like `0` or `False`.  Probably not ideal.
2. I would advocate for this validation function to purely evaluate actual aliases and to leave the acceptance of `None` or other allowed empty values to happen at a higher level.  Same with `validate_ethpm_dir`

tests/core/test_validation.py
```diff
@@ -0,0 +1,91 @@
...
+
+from ethpm_cli.exceptions import InstallError, UriNotSupportedError, ValidationError
+from ethpm_cli.validation import validate_cli_args
+
+
+@pytest.fixture
+def args():
+    namespace = Namespace()
+    setattr(namespace, "uri", "ipfs://QmbeVyFLSuEUxiXKwSsEjef6icpdTdA4kGG9BcrJXKNKUW")
+    setattr(namespace, "ethpm_dir", None)
```
pipermerriam: Any reason you are using `setattr` here instead of just assigning to the attribute?

tests/core/test_validation.py
```diff
@@ -0,0 +1,91 @@
...
+    ethpm_dir = p / "ethpm_packages"
+    ethpm_dir.mkdir()
+    monkeypatch.chdir(p)
+    setattr(args, "ethpm_dir", Path("./ethpm_packages"))
+
+    assert validate_cli_args(args) is None
+
+
+def test_validate_cli_args_rejects_invalid_absolute_paths(args, tmpdir):
+    invalid_path = Path(tmpdir) / "ethpm_packages"
```
pipermerriam: what is it that makes this invalid?

pipermerriam (approved):
I *think* this might come in handy for testing this as a black box:

https://pexpect.readthedocs.io/en/stable/

tests/core/test_validation.py
njgheorghita: It's not clear, but the fact that the directory doesn't exist. I updated to make it a little more obvious

PR #3, opened 2019-04-29
Scrape w3 for ipfs uris and write them to disk

njgheorghita:
## What was wrong?
CLI command needed to scrape blockchain for `VersionRelease` events - grab version release data - and write any emitted ipfs assets to disk

## How was it fixed?
Wrote `Scraper` class that takes a web3 instance and scrapes for `VersionRelease` data, writing any found data to disk.

#### Cute Animal Picture
image


ethpm_cli/scraper.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def __init__(
+        self,
+        w3: Web3,
+        ws_w3: Web3 = None,
+        content_dir: Path = None,
+        logger: Logger = None,
+    ) -> None:
+        self.w3 = w3
+        # ws w3 instance needed since infura requires ws for certain jsonrpc calls
```
njgheorghita: @kclowes - Infura only supports [certain jsonrpc](https://infura.io/docs) endpoints over ws (i.e. `eth_newFilter`) and others only over https (i.e. `eth_getBlock`).  I had a thought that it might be worth making web3 smart enough to direct the right calls to the relevant endpoints - and avoid having to use 2 different w3 instances - but it only applies to infura, so I'm not sure if it's worth it, just a thought.

ethpm_cli/constants.py
```diff
@@ -1 +1,18 @@
+import json
+
+from eth_utils import keccak, to_hex
+from ethpm.constants import INFURA_API_KEY
+
+CONTENT_DIR = "ipfs_assets"
```
pipermerriam: I think that this should make use of the [XDG spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) which would mean storing this data somewhere like `$HOME/.local/share/ethpm` or maybe `$HOME/.local/cache/ethpm` depending on the semantics of how we want to store this data.

You can see how we negotiate these directories [here in the trinity codebase](https://github.com/ethereum/trinity/blob/master/trinity/_utils/xdg.py)

*sidenote* I think that file could easily be extracted to a small python library to make it easier to generate those paths for different applications.

ethpm_cli/scraper.py
```diff
@@ -0,0 +1,195 @@
...
+    Scrape blockchain for all valid 'VersionRelease' events
+    Resolves manifest uris & nested uris to a json object
+    """
+
+    def __init__(
+        self,
+        w3: Web3,
+        ws_w3: Web3 = None,
+        content_dir: Path = None,
+        logger: Logger = None,
```
pipermerriam: I've come to see passing loggers around as an anti-pattern.  I'd recommend just setting it on the class.

```python
class Scraper:
    logger = logging.getLogger('ethpm.cli.scraper.Scraper')
```

ethpm_cli/scraper.py
```diff
@@ -0,0 +1,195 @@
...
+    ) -> None:
+        self.w3 = w3
+        # websocket w3 instance needed since infura requires ws for certain jsonrpc calls
```
pipermerriam: Take a look at this: https://github.com/ethereum/web3.py/blob/71b8bf03dc6d332dd97d8902a38ffab6f8b5a5ab/web3/middleware/filter.py#L284

This PR's functionality seems like the ideal use case for that middleware.

ethpm_cli/scraper.py
```diff
@@ -0,0 +1,195 @@
...
+        self.ws_w3 = ws_w3 or w3
+        self.version_release_data: OrderedDict[str, Any] = OrderedDict()
+        self.last_processed_block = 0
```
pipermerriam: This statefulness gets bugs me a little, but I think that's mostly because I can see how we could do some small changes to remove it, but that may mean giving up on the scraper class, or at least extracting the functionality out of it.  Here's the gist.

```
class Config(NamedTuple):
    w3: Web3
    from_block: identifier  # maybe `None` signifies genesis
    to_block: identifier  # maybe `None` signifies to current


def process_uri(config, ipfs_uri) -> Set[Tuple[URI, Manifest]]:
    # download -> validate -> recurse-dependencies
    ...


def scrape_manifests_from_chain(config):
    # Low level function that is somewhat naive of statefulness of system.
    filter = w3.eth.filter(...)

    for log in filter:
        uri = get_uri_from_log(log)
        yield log, process_uri(config, uri)


def scrape(config, to_path: Path):
    # High level function that manages statefullness of system
    try:
        for log, manifests in scrape_chain(self.config):
            ...  
            # maybe write them to disk?
    except ...:
        # maybe save our progress (last processed log...)
```

pipermerriam (commented):
Incomplete review, will finish it out in a bit.

ethpm_cli/scraper.py
```diff
@@ -0,0 +1,195 @@
...
+        if version_release_logs:
+            formatted_logs = format_version_release_logs(version_release_logs)
+            self.version_release_data = assoc(
+                self.version_release_data,
+                str(self.last_processed_block),
+                formatted_logs,
+            )
+
+        self.logger.info(
+            f"Block #:{self.last_processed_block} scraped. "
```
pipermerriam: see https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#formatting-log-messages

pipermerriam (commented):
Going to wait to review the rest until later but wanted to ask for a more detailed write-up of what it scrapes and what data it writes to disk, and how that data is organized.

ethpm_cli/scraper.py
njgheorghita: @kclowes scratch that re: [this](https://github.com/ethpm/ethpm-cli/pull/3#discussion_r279393654) `local_filter_middleware` works great!

ethpm_cli/scraper.py
```diff
@@ -0,0 +1,195 @@
...
+from typing import Any, Dict, Iterable, List, Set, Tuple  # noqa: F401
+
+from eth_utils import to_dict, to_hex, to_int, to_list
+from eth_utils.toolz import assoc
+from ethpm.typing import URI, Address
+from ethpm.utils.backend import resolve_uri_contents
+from ethpm.utils.ipfs import extract_ipfs_path_from_uri, is_ipfs_uri
+from ethpm.utils.uri import is_supported_content_addressed_uri
+from web3 import Web3
+from web3._utils.events import get_event_data
```
pipermerriam: Noting the import from a private API.  IIUC I think you should be able to define a contract factory that has the relevant event ABI and then use the contract APIs to pull logs that are properly parsed.

njgheorghita:
@pipermerriam Here's a little more on the structure of the data stored, it's more or less just the minimum I thought I needed to get things working
https://gist.github.com/njgheorghita/1ed9e6e061a5a84354cf08bc8d32812b

pipermerriam (commented):
What would you say to dropping the complexity here a bit.  My main goal is to reduce the amount of data stored in JSON documents to be constant size (or very near constant).

Instead of storing all of the metadata, what if we only store the following:

- what blocks we've scraped either in the form of a singular range, or in the form of a list of ranges which we collapse when possible.
- a directory (or nested set of directories) where the contents can be stored and looked up by hash:
  - Something like `$XDG_CACHE/ethpm/ipfs/<first-two-hex-bytes-of-hash>/<second-two-hex-bytes-of-hash>/<remainder-of-hash>/manifest.json`
  -  Maybe store the metadata about the registry, package names, version, whatever alongside the `manifest.json`.

Basically trying to have the data be as self-describing as possible so that another tool that wants a file can look for it without having to understand the whole schema of the various JSON documents that hold metadata.  Also, trying to avoid having **any** large JSON documents.

njgheorghita:
Updated ipfs assets dir structure 
```
- 'ipfs/'
    - 'chain_data.json'
    - 'Qm/'
        - 'be/'
            - ...
        - 'dv/'  (dir with first 2 unique bytes)
            - 'ZEW3AaUntDfFkcbdnYzeLAAeD4YFeixQsdmHF88T6Q/' (dir with remainder of hash)
                - 'QmdvZEW3AaUntDfFkcbdnYzeLAAeD4YFeixQsdmHF88T6Q' (file w/ name of entire hash, stores contents)
```
@pipermerriam  Thoughts on this structure? I eliminated storing metadata since it's not obviously necessary and tracking all the assets became fairly complex since we're not only storing manifests, but also their build dependencies and any contract files, documentation etc.. But, I can re-implement it if you think it's going to be important.

I've added a list of `"registry_addresses"` to the `"chain_data.json"` file. It violates your "keep json small" rule, but it shouldn't grow too quickly, and seems like that information could be useful. (This list of "active EthPM V2" registry addresses was the prior motivation for storing any metadata in the first place)

pipermerriam:
What about separating the scraping of registry addresses into a separate utility?  Feel free to punt on that and keep things as-is for now if it's useful to you.

For directory structure I'd go one level deeper with `Qm/dv/ZE/QMdvZEW3.....`  The extra level of namespacing is cheap and it is a huge PITA when you get a directory with way too many things in it because at a certain point `ls` stops working.

As for metadata, are we still scraping the files, just not storing information about what file is what?  As long as we're pulling the contract source files and dependency manifests I think we're :+1: .  Not pulling dependencies is probably a no-go since without the dependencies a package is likely useless.

njgheorghita:
> What about separating the scraping of registry addresses into a separate utility? Feel free to punt on that and keep things as-is for now if it's useful to you.

Sounds good - I'll open up a new pr to handle that task. 

> For directory structure I'd go one level deeper with Qm/dv/ZE/QMdvZEW3..... 

👍 

> As long as we're pulling the contract source files and dependency manifests I think we're 👍 . 

We're 👍

pipermerriam:
giving one last :eyes:

__main__.py
```diff
@@ -21,10 +28,40 @@ def get_logger():
...
+    w3 = Web3(load_provider_from_uri(INFURA_HTTP_URI))
+    w3.middleware_onion.add(local_filter_middleware)
+    return w3
+
+
+def scraper(args):
+    w3 = setup_scraper()
+    ethpmcli_dir = get_xdg_ethpmcli_root()
+    scrape(w3, ethpmcli_dir, args.start_block)
+    logger.info("All blocks scraped up to # %d.", w3.eth.getBlock)
```
pipermerriam: Maybe have scrape return the latest scraped block or even some sort of `ScrapeResult` data object to prevent a race condition where a new block comes in between `scrape` finishing and this reporting at which point the user might later use this log statement to determine which block to start from and miss any intermediate blocks.  Also, probably good to ensure the statement has both block number and hash (current statement looks like it dumps the whole block which seems like it'd be noisy)

__main__.py
```diff
@@ -53,25 +90,28 @@ def parse_arguments():
...
         install_package(package, config)
         logger.info(
-            f"{package.alias} package sourced from {args.uri} "
-            f"installed to {config.ethpm_dir}."
+            "%s package sourced from %s installed to %s.",
+            package.alias,
+            args.uri,
+            config.ethpm_dir,
         )
+    if args.command == "scrape":
```
pipermerriam, line 112: Consider using this pattern (can be done in another PR).

https://github.com/ethereum/trinity/blob/9608a54c9985f60a32278b3115754f5fadf0cb6e/trinity/plugins/builtin/fix_unclean_shutdown/plugin.py#L28-L33

https://github.com/ethereum/trinity/blob/9608a54c9985f60a32278b3115754f5fadf0cb6e/trinity/bootstrap.py#L196-L197

Each cli command specifies it's own execution function so you can keep this area of the codebase clean of the individual business logic for each CLI command.

__main__.py
```diff
@@ -53,25 +90,28 @@ def parse_arguments():
...
+        scraper(args)
     else:
         logger.info(
-            f"{args.command} is an invalid command. Use `ethpmcli --help` "
-            "to see the list of available commands."
+            "%s is an invalid command. Use `ethpmcli --help` to see the list of available commands.",
+            args.command,
         )
```
pipermerriam, line 118: This should probably be done via an [parser.error](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.error) API so that it exits with a non-zero exit code to ensure scripts don't interpret this as a happy path.

ethpm_cli/_utils/various.py
```diff
@@ -0,0 +1,13 @@
+import collections
+from typing import Any, Iterable, List
+
+from eth_utils import to_list
+
+
+@to_list
+def flatten(l: List[Any]) -> Iterable[Any]:
```
pipermerriam, line 8: Curios if mypy allows this:

```
TValue = TypeVar("TValue")
TList = Sequence[Union[TValue, "TList"]]  # recursive definition...
```

This would in theory allow retention of the value type.  Not a big deal if it doesn't work.

ethpm_cli/_utils/xdg.py
```diff
@@ -0,0 +1,42 @@
+import os
+from pathlib import Path
+
+from ethpm_cli.exceptions import AmbigiousFileSystem
+
+
+def get_home() -> Path:
+    try:
+        return Path(os.environ["HOME"])
```
pipermerriam: Should probably use [`Path.home`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.home)

ethpm_cli/scraper.py
```diff
@@ -0,0 +1,215 @@
...
+import json
+import logging
+from pathlib import Path
+from typing import Any, Dict, Iterable, List, Set, Tuple  # noqa: F401
+
+from eth_utils import to_dict, to_list
+from eth_utils.toolz import assoc
+from ethpm.typing import URI, Address
+from ethpm.utils.backend import resolve_uri_contents
+from ethpm.utils.ipfs import extract_ipfs_path_from_uri, is_ipfs_uri
```
pipermerriam, line 11: These need to eventually be exposed from the `ethpm` library as a public API.  Currently this is importing from private API namespace.

ethpm_cli/scraper.py
```diff
@@ -0,0 +1,215 @@
...
+    try:
+        active_block = import_ethpm_dir(ethpm_dir, w3, start_block)
+        scraped_manifests = scrape_block_for_manifests(w3, active_block)
+
+        update_chain_data(chain_data_path, active_block, scraped_manifests)
+        write_ipfs_uris_to_disk(ethpm_dir, scraped_manifests)
+
+    except BlockNotFoundError:
+        logger.info("No more blocks available.")
+        return None
```
pipermerriam: I *think* convention is just a bare `return` since we aren't actually intending to return the `None` value, just that we're returning.

ethpm_cli/scraper.py
```diff
@@ -0,0 +1,215 @@
...
+
+logger = logging.getLogger("ethpm_cli.scraper.Scraper")
+
+
+def scrape(w3: Web3, ethpm_dir: Path, start_block: int = 0) -> None:
+    """
+    Scrapes VersionRelease event data starting from start_block.
+    """
+    chain_data_path = ethpm_dir / "chain_data.json"
+    try:
```
pipermerriam: This `try` block looks like it contains more statements than necessary.  Can `BlockNotFoundError` come from any of these four statements?  If not, we should adjust to have only the statements which can raise be present in the `try` (even if you need to make more than one `try/except` block.

ethpm_cli/scraper.py
```diff
@@ -0,0 +1,215 @@
...
+        write_ipfs_uris_to_disk(ethpm_dir, scraped_manifests)
+
+    except BlockNotFoundError:
+        logger.info("No more blocks available.")
+        return None
+
+    except BlockAlreadyScrapedError:
+        active_block = start_block + 1
+
+    scrape(w3, ethpm_dir, active_block)
```
pipermerriam: This recursive approach is going to be problematic.  It will add a level of recursion for every already scraped block which will VERY quickly exceed the stack depth limit.  Should adjust this to use a loop and a counter and to break from the loop and enter the *real* scrape function once it finds the first unscraped block.

ethpm_cli/scraper.py
```diff
@@ -0,0 +1,215 @@
...
+    If no ethpm_dir found, creates an ethpm_dir and returns a 0.
+    """
+    if ethpm_dir.is_dir():
+        chain_data_path = ethpm_dir / "chain_data.json"
+        validate_chain_data_store(chain_data_path, w3)
+
+        if start_block >= w3.eth.blockNumber:
+            raise BlockNotFoundError(
+                f"Block number: {start_block} not available on provided web3 "
+                f"instance with latest block number of {w3.eth.blockNumber}."
```
pipermerriam: race condition here,  probably should pull off `w3.eth.blockNumber` to a local variable so that you don't encounter a situation where it changes between the two locations this property is accessed.

ethpm_cli/scraper.py
```diff
@@ -0,0 +1,215 @@
...
+        (ethpm_dir / "Qm").mkdir()
+        chain_data_path = ethpm_dir / "chain_data.json"
+        chain_data_path.touch()
+        init_json = {
+            "chain_id": w3.eth.chainId,
+            "registry_addresses": [],
+            "scraped_blocks": [{"min": "0", "max": "0"}],
+        }
+        chain_data_path.write_text(f"{json.dumps(init_json, indent=4)}\n")
+        return 1
```
pipermerriam: This conflates the two concepts of:

-  *ensure that the directory and file structure is present*
- *tell me which block I should start scraping from*

I recommend separating these into two different functions.

ethpm_cli/scraper.py
```diff
@@ -0,0 +1,215 @@
...
+    chain_data_updated_addrs = assoc(
+        chain_data_updated_blocks, "registry_addresses", updated_addrs
+    )
+    chain_data_path.write_text(f"{json.dumps(chain_data_updated_addrs, indent=4)}\n")
+
+
+@to_list
+def blocks_to_ranges(blocks_list: List[int]) -> Iterable[Dict[str, str]]:
+    """
+    Takes a list of block numbers, and returns them grouped into min/max ranges.
```
pipermerriam, line 90: I *think* I understand what this says but I'm not sure.  Can you maybe provide a small code example in this docstring so it is illustrative?

Is this what it does?

https://github.com/ethereum/trinity/pull/569/files#diff-3c5a1b448f91c1258a0e43b661b3af54R16

ethpm_cli/scraper.py
```diff
@@ -0,0 +1,215 @@
...
+) -> None:
+    ipfs_dir = ethpm_dir / "Qm"
+    all_manifest_uris = [
+        version_release_data["manifestURI"]
+        for version_release_data in manifests.values()
+        if is_supported_content_addressed_uri(version_release_data["manifestURI"])
+    ]
+    base_ipfs_uris = [uri for uri in all_manifest_uris if is_ipfs_uri(uri)]
+    nested_ipfs_uris = [pluck_ipfs_uris_from_manifest(uri) for uri in all_manifest_uris]
+    all_ipfs_uris = set(flatten(nested_ipfs_uris) + base_ipfs_uris)
```
pipermerriam, line 117: This might be more appropriate done as `set(flatten(nested_ipfs_uris)).update(*base_ipfs_uris)`

ethpm_cli/scraper.py
```diff
@@ -0,0 +1,215 @@
...
+    scraped_ranges = [
+        list(range(int(rnge["min"]), int(rnge["max"]) + 1)) for rnge in scraped_blocks
+    ]
+    return flatten(scraped_ranges)
+
+
+def write_ipfs_uris_to_disk(
+    ethpm_dir: Path, manifests: Dict[Address, Dict[str, str]]
+) -> None:
+    ipfs_dir = ethpm_dir / "Qm"
```
pipermerriam: I don't think this is ok.  While every IPFS URI you've worked with starts with this, it isn't guaranteed.  This prefix merely specifies the hash type and all implementations use SHA256.   Hard coding `Qm` is going to shoot us in the foot when an alternate hash-scheme shows up.

ethpm_cli/scraper.py
```diff
@@ -0,0 +1,215 @@
...
+
+        if not unique_hash_prefix_dir.is_dir():
+            unique_hash_prefix_dir.mkdir()
+
+        if not remainder_hash_dir.is_dir():
+            remainder_hash_dir.mkdir()
+            asset_dest_path = remainder_hash_dir / ipfs_hash
+            asset_dest_path.touch()
+            asset_dest_path.write_bytes(resolve_uri_contents(uri))
+            logger.info("%s written to disk.", uri)
```
pipermerriam: Probably prudent to include the `path` in this message too..

tests/core/ipfs/Qm/NQ/euwMDGJ7UiLaRjwzAoekcaKLp9TjiqeFdovj3syN1n/QmNQeuwMDGJ7UiLaRjwzAoekcaKLp9TjiqeFdovj3syN1n
```diff
@@ -0,0 +1,23 @@
+pragma solidity ^0.4.24;
```
pipermerriam: Looks like fixtures that should be zipped up to prevent accidentally modifying them in-place?

pipermerriam (approved):
Feel free to move this forward at whatever pace you need.  Few things in here that should be cleaned up but it's fine to punt them into tracking issues if you so desire.

ethpm_cli/_utils/various.py
njgheorghita: Not supported *yet* ...
`Recursive types not fully supported yet, nested types replaced with "Any"`

ethpm_cli/scraper.py
njgheorghita: yup! more or less, just formatted slightly differently

PR #16, opened 2019-05-15
Intelligently find start block

### LINKED ISSUE #9: Intelligent selection of start block
## What was wrong?

There are a few million blocks at the start of the mainnet chain that can be reasonably skipped since we know they came into existence **before** the ethpm spec existed.

These blocks should be safe to skip when scraping the chain.

## How can it be fixed?

Ideally we want to *allow* these blocks to be scraped but skip them by default.

- `ethpm-cli scrape --start-block=auto` (new fancy behavior)
- `ethpm-cli scrape --start-block=1234`: starts at block `1234`

The `auto` should do a binary search of the chain checking the block timestamps until it finds the first block *before* whatever spec was published which contains the event signature we are scraping for.  We should reliably get to start scraping at that block.

### LINKED ISSUE #12: Scrape blocks in batches
## What was wrong?

The current scraper scrapes each block individually.  This is very slow.

## How can it be fixed?

Each block is scraped by using a `getLogs` call for that singular block.  This should be easy enough to scale up to batching across multiple blocks to reduce the total number of needed network requests.

njgheorghita:
## What was wrong?
- Scrape blocks in batches: Fixes #12 
- Intelligent selection of start block: Fixes #9 

#### Cute Animal Picture
image

### DIFF
--- a/ethpm_cli/main.py
+++ b/ethpm_cli/main.py
@@ -3,6 +3,7 @@ import logging
 from pathlib import Path
 import sys
 
+from eth_utils import to_hex
 import pkg_resources
 from web3 import Web3
 from web3.middleware import local_filter_middleware
@@ -43,7 +44,7 @@ def scraper(args: argparse.Namespace) -> None:
     logger.info(
         "All blocks scraped up to # %d: %s.",
         last_scraped_block,
-        last_scraped_block_hash,
+        to_hex(last_scraped_block_hash),
     )
 
 
--- a/ethpm_cli/main.py
+++ b/ethpm_cli/main.py
@@ -3,7 +3,7 @@ import logging
 from pathlib import Path
 import sys
 
-from eth_utils import to_hex
+from eth_utils import humanize_hash
 import pkg_resources
 from web3 import Web3
 from web3.middleware import local_filter_middleware
@@ -44,7 +44,12 @@ def scraper(args: argparse.Namespace) -> None:
     logger.info(
         "All blocks scraped up to # %d: %s.",
         last_scraped_block,
-        to_hex(last_scraped_block_hash),
+        humanize_hash(last_scraped_block_hash),
+    )
+    logger.debug(
+        "All blocks scraped up to # %d: %s.",
+        last_scraped_block,
+        last_scraped_block_hash,
     )
 
 
--- a/ethpm_cli/scraper.py
+++ b/ethpm_cli/scraper.py
@@ -20,6 +20,8 @@ from ethpm_cli.validation import validate_chain_data_store
 logger = logging.getLogger("ethpm_cli.scraper.Scraper")
 
 
+BLOCK_INTERVAL = 50
+
 def scrape(w3: Web3, ethpm_dir: Path, start_block: int = 0) -> int:
     """
     Scrapes VersionRelease event data starting from start_block.
@@ -39,22 +41,27 @@ def scrape(w3: Web3, ethpm_dir: Path, start_block: int = 0) -> int:
 
     ethpmdir_block = max(get_scraped_blocks(chain_data_path))
     active_block = start_block if start_block else ethpmdir_block
-    for block in range(active_block, latest_block):
-        try:
-            validate_unscraped_block(block, chain_data_path)
-        except BlockAlreadyScrapedError:
-            pass
+
+    for from_block in range(active_block, latest_block, BLOCK_INTERVAL):
+        if (from_block + BLOCK_INTERVAL) > latest_block:
+            to_block = latest_block
         else:
-            scraped_manifests = scrape_block_for_manifests(w3, block)
-            update_chain_data(chain_data_path, block, scraped_manifests)
+            to_block = from_block + BLOCK_INTERVAL
+
+        if block_range_needs_scraping(from_block, to_block, chain_data_path):
+            scraped_manifests = scrape_block_range_for_manifests(w3, from_block, to_block)
+            update_chain_data(chain_data_path, from_block, to_block, scraped_manifests)
             write_ipfs_uris_to_disk(ethpm_dir, scraped_manifests)
+
     return latest_block
 
 
-def validate_unscraped_block(block: int, chain_data_path: Path) -> None:
+def block_range_needs_scraping(from_block: int, to_block: int, chain_data_path: Path) -> bool:
     all_scraped_blocks = get_scraped_blocks(chain_data_path)
-    if block in all_scraped_blocks:
-        raise BlockAlreadyScrapedError(f"Skipping block #{block}. Already processed.")
+    for block in range(from_block, to_block):
+        if block not in all_scraped_blocks:
+            return True
+    return False
 
 
 def initialize_ethpm_dir(ethpm_dir: Path, w3: Web3) -> None:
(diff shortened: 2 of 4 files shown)


ethpm_cli/main.py
```diff
@@ -43,7 +44,7 @@ def scraper(args: argparse.Namespace) -> None:
     logger.info(
         "All blocks scraped up to # %d: %s.",
         last_scraped_block,
-        last_scraped_block_hash,
+        to_hex(last_scraped_block_hash),
```
pipermerriam: If you want you can use `humanize_hash` here which makes long hashes more human readable at the cost of omitting the middle bytes.

You can remedy this by issueing a second logging statement at `DEBUG` level that contains the full hash.

ethpm_cli/scraper.py
```diff
@@ -14,18 +15,23 @@
...
 from ethpm_cli._utils.various import flatten
 from ethpm_cli.constants import VERSION_RELEASE_ABI
-from ethpm_cli.exceptions import BlockAlreadyScrapedError, BlockNotFoundError
+from ethpm_cli.exceptions import BlockNotFoundError
 from ethpm_cli.validation import validate_chain_data_store
 
 logger = logging.getLogger("ethpm_cli.scraper.Scraper")
 
+# https://github.com/ethereum/EIPs/commit/123b7267b6270914a822001c119d11607e695517
+VERSION_RELEASE_TIMESTAMP = 1_552_564_800  # March 14, 2019
```
pipermerriam, line 26: mindblown

HOW DID I NOT KNOW YOU COULD DO THIS....

ethpm_cli/scraper.py
```diff
@@ -37,24 +43,57 @@ def scrape(w3: Web3, ethpm_dir: Path, start_block: int = 0) -> int:
...
+            update_chain_data(chain_data_path, from_block, to_block, scraped_manifests)
             write_ipfs_uris_to_disk(ethpm_dir, scraped_manifests)
+        else:
+            logger.info("Block range: %d - %d already scraped.", from_block, to_block)
+
     return latest_block
 
 
-def validate_unscraped_block(block: int, chain_data_path: Path) -> None:
+def get_ethpm_birth_block(w3: Web3, from_block: int, to_block: int) -> int:
```
pipermerriam: You'll have an easier time testing this function if you make it generic so you can specify the target timestamp as well.  That should allow you to pretty easily generate a short testing chain using the [`time_travel` functionality ]*https://github.com/ethereum/eth-tester#time-travel) of `eth-tester` and verify correctness.

ethpm_cli/scraper.py
```diff
@@ -37,24 +43,57 @@ def scrape(w3: Web3, ethpm_dir: Path, start_block: int = 0) -> int:
...
+    version_release_date = datetime.fromtimestamp(VERSION_RELEASE_TIMESTAMP)
+    from_date = datetime.fromtimestamp(w3.eth.getBlock(from_block)["timestamp"])
+    delta = version_release_date - from_date
+
+    if delta.days == 0 and from_date < version_release_date:
+        return from_block
+
+    elif from_date < version_release_date:
+        updated_block = int((from_block + to_block) / 2)
+        return get_ethpm_birth_block(w3, updated_block, to_block)
```
pipermerriam: I'm only giving you half credit for a recursive implementation.  It's fine to leave as-is but would you mind opening an issue indicating this should be refactored to be non-recursive at some point (probably a good issue to tag with "Good First Issue")

ethpm_cli/scraper.py
```diff
@@ -14,18 +15,23 @@
...
-from ethpm_cli.exceptions import BlockAlreadyScrapedError, BlockNotFoundError
+from ethpm_cli.exceptions import BlockNotFoundError
 from ethpm_cli.validation import validate_chain_data_store
 
 logger = logging.getLogger("ethpm_cli.scraper.Scraper")
 
+# https://github.com/ethereum/EIPs/commit/123b7267b6270914a822001c119d11607e695517
+VERSION_RELEASE_TIMESTAMP = 1_552_564_800  # March 14, 2019
+
+BLOCK_INTERVAL = 5000
```
pipermerriam: I think `batch` size might be a more semantically meaningful name.

ethpm_cli/scraper.py
```diff
@@ -73,12 +112,16 @@ def initialize_ethpm_dir(ethpm_dir: Path, w3: Web3) -> None:
...
     chain_data = json.loads(chain_data_path.read_text())
 
-    all_scraped_blocks = get_scraped_blocks(chain_data_path)
-    updated_blocks = blocks_to_ranges(set(all_scraped_blocks + [block_number]))
+    old_scraped_blocks = get_scraped_blocks(chain_data_path)
+    new_scraped_blocks = list(range(from_block, to_block))
+    updated_blocks = blocks_to_ranges(set(old_scraped_blocks + new_scraped_blocks))
 
     chain_data_updated_blocks = assoc(chain_data, "scraped_blocks", updated_blocks)
     chain_data_path.write_text(f"{json.dumps(chain_data_updated_blocks, indent=4)}\n")
```
pipermerriam: you might consider using the tempdir/tempfile approach for this as well to ensure that you don't end up with a corrupt file if the process dies half-way through writing.

- copy file
- update the copy
- replace original file with updated copy

ethpm_cli/scraper.py
```diff
@@ -198,14 +242,23 @@ def process_entries(
 ) -> Iterable[Tuple[str, str]]:
     for entry in all_entries:
         if entry["address"] == address:
+            logger.info(
+                "<Package %s==%s> released on registry @ %s.\n" "Manifest URI: %s\n",
```
pipermerriam, line 265: *nitpick*.  Why the newline in the middle?

ethpm_cli/scraper.py
```diff
@@ -198,14 +242,23 @@ def process_entries(
...
+                entry["args"]["packageName"],
+                entry["args"]["version"],
+                address,
```
pipermerriam, line 268: Doesn't have to be for this PR but doing a reverse resolution on this address to see if it is an ENS domain would be really good UX (still should probably display the address, but you could *humanize* it and truncate the middle in that case)

pipermerriam (approved):

ethpm_cli/main.py
njgheorghita: Ohh that's cool, didn't know about `humanize_hash`

ethpm_cli/scraper.py
njgheorghita: Yeah, it's lovely - 100% credit to `black` for pointing it out

ethpm_cli/scraper.py
njgheorghita: Yeah, no problem. I'm starting to sense that you don't think recursive solutions are the right tool to reach for most of the time?

ethpm_cli/scraper.py
njgheorghita: Just aesthetics. The entire string"<Package .... URI: %s\n." just seemed a little long & too much data for one line. But I don't have strong feelings about it. Do you think it's better off as a one-liner?

ethpm_cli/scraper.py
pipermerriam: Nah, for this the recursion limit is very unlikely to be hit, but that's assuming that the program executes correctly.  I've re-implemented binary searches a bunch of time and plenty of them ended up with bugs that resulted in an infinite loop which would have translated to an error due to stack depth here.   Neither of those are really better necessarily.

Really it is that recursion is a sort of code smell.  

- It costs extra in some ways because python has to allocate more call frames.
- You have to know the bounds of your algorithm to evaluate if stack limit will be a problem.
- other things that probably show up if you google "why is recursion bad"

Maybe the right thing would be *"Recursion is by default the wrong tool, and is only the right tool if it has been closely evaluated in the context of the problem it solves".

ethpm_cli/scraper.py
pipermerriam: I'd be inclined to stick it on one line.  Lets you easily do `wc -l` on the output to count how many packages were found?  I'm sure there are good reasons for either though.

PR #17, opened 2019-05-15
Uninstall & list packages

### LINKED ISSUE #13: Uninstall packages from ethpmdir
The ability to uninstall packages / specific versions could be useful. Base `ethpm.lock` must be updated when a package is uninstalled.

Also potentially useful - some kind of `list` command to list all installed pkgs.

njgheorghita:
## What was wrong?
- List installed pkgs & dependencies: Fixes #13 
- Uninstall packages: Fixes #13

#### Cute Animal Picture
image


ethpm_cli/install.py
```diff
@@ -47,11 +54,39 @@ def install_package(pkg: Package, config: Config) -> None:
...
+
+
+def list_installed_packages(config: Config) -> None:
+    for pkg_data in get_installed_packages(config.ethpm_dir):
+        logger.info(pkg_data)
+
+
+@to_list
+def get_installed_packages(ethpm_dir: Path) -> Iterable[str]:
+    installed_pkgs = reversed(sorted(ethpm_dir.glob("**/manifest.json")))
```
pipermerriam: I don't think this is a safe mechanism to use.  A package could maliciously ship a file called `manifest.json` which would be caught by this mechanism and corrupt the displayed data.

ethpm_cli/install.py
```diff
@@ -47,11 +54,39 @@ def install_package(pkg: Package, config: Config) -> None:
...
+    for manifest_path in installed_pkgs:
+        manifest = json.loads(manifest_path.read_text())
+        num_deep = str(manifest_path.relative_to(ethpm_dir)).count("/")
+        yield f"{num_deep * '--'} <Package {manifest['package_name']}=={manifest['version']}>"
```
pipermerriam: While this is fine and doesn't absolutely need to be addressed, I think you will want to restructure these function a little.

- `get_installed_package_tree`: Returns a data structure which represents the full tree of installed packages.
- A mechanism for displaying that information.

Here's an idea that might work well.

```python
class InstalledPackageTree(NamedTuple):
    depth: int
    path: Path
    manifest: ...
    children: Tuple["InstalledPackageTree", ...]  # self referencing, mypy might not like this but data structure wise this should be fine.
```

You should be able to use this data structure to store the data about the package tree.

Next, a recursive implementation that generates the data structure.

For each of the top level dependencies, populate them using the following function.

```python
def get_installed_package_tree(base_dir, depth=0):
    manifest = ...  # the loaded manifest (either raw JSON or the parsed internal representation)
    dependency_dirs = ...  # the base directories for each of this packages installed dependencies
    children = tuple(
        get_installed_package_tree(dependency_dir, depth + 1)
        for dependency_dir in dependency_dirs
    )
    return InstalledPackageTee(depth, base_dir, manifest, children)
```

You should then have a tuple/iterable of `InstalledPackageTree` objects that represent all of the installed packages.

Next to display the data, we can just implement a method on the `InstalledPackageTree`.  This could be `__str__` or something more formal like `format_for_display()`.

```python
def __str__(self) -> str:
    prefix = "  " * depth
    if self.children:
        children = "\n" + "\n".join((str(child) for child in self.children))
    else:
        children = ""
    return (
        f"{prefix}{self.package_identifier}=={self.package_version} ({self.content_hash_as_hex}){children}"
    )
```

Then your function to display these is just to log the `__str__` representation of each of the top level `InstalledPackageTree` objects to the terminal.

ethpm_cli/install.py
```diff
@@ -47,11 +54,39 @@ def install_package(pkg: Package, config: Config) -> None:
...
+
+
+def uninstall_package(package_name: str, config: Config) -> None:
+    if not is_package_installed(package_name, config):
+        raise InstallError(
+            f"Unable to uninstall {package_name} from {config.ethpm_dir}"
+        )
+
+    shutil.rmtree(config.ethpm_dir / package_name)
+    uninstall_from_ethpm_lock(package_name, (config.ethpm_dir / "ethpm.lock"))
```
pipermerriam: So *if* the process gets killed between these execution of these two statements then you will still have the package in the lockfile but the actual package files will be missing.

I'd recommend making this atomic by:

- do a full recursive copy of the `ethpm_dir` to a temporary directory.
- delete the package in the temp dir
- update the lock in the tempdir
- replace the original directory with the temporary directory.

The operation is only complete once the last operation occurs and any failure before then will leave the data in the same state it started.

ethpm_cli/main.py
```diff
@@ -91,6 +92,15 @@ def parse_arguments() -> argparse.ArgumentParser:
         action="store_true",
         help="Flag to use locally running IPFS node.",
     )
+
+    list_parser = subparsers.add_parser("list", help="List installed packages")
+    list_parser.add_argument(
```
pipermerriam: I would recommend moving the definition of your parser out of this function body either into the top level of this module or even into it's own module to 1. reduce the size of this function body and 2. maybe reduce the size of this module.

tests/core/test_install.py
```diff
@@ -109,3 +119,35 @@ def test_install_multiple_packges(config, test_assets_dir, owned_pkg):
...
+        uninstall_package("invalid", config)
+
+
+def test_list(config, owned_pkg, wallet_pkg, caplog):
+    install_package(owned_pkg, config)
+    install_package(wallet_pkg, config)
+
+    with caplog.at_level(logging.INFO):
+        list_installed_packages(config)
+        assert "-- <Package wallet==1.0.0>" in caplog.text
```
pipermerriam: You might add the newline at the end of this text to be more precise in verifying these are properly formatted since the current format would allow them all munged on the same line.

pipermerriam (changes requested):

njgheorghita:
@piper ping for more :eyes:

also, re: the failing circle builds ... Is it possible that the bug - which only recently appeared - could be on Infura's end? Or is it more likely on my end, and I have to figure out how to install/configure the proper verification in the build
 
```SSLError(MaxRetryError("HTTPSConnectionPool(host='ipfs.infura.io', port=5001): Max retries exceeded with url: /api/v0/...[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1056)')))"))```

**Comment by pipermerriam** [comment]
Likely on our/your end.  Have you done some googling of that error?  That's where I would start.

**Comment by pipermerriam** [comment]
In cases where you want additional review it's nice to not smush your commits until the very end when you merge.  Otherwise it's hard to see what changed from my end.

File: ethpm_cli/install.py [hunk 6]
```diff
@@ -47,11 +34,97 @@ def install_package(pkg: Package, config: Config) -> None:
...
-    update_ethpm_lock(pkg, (config.ethpm_dir / "ethpm.lock"))
+    install_to_ethpm_lock(pkg, (config.ethpm_dir / "ethpm.lock"))
+
+
+_InstalledPackageTree = namedtuple(
+    "InstalledPackageTree", "depth path manifest children content_hash"
+)
+
+
+class InstalledPackageTree(_InstalledPackageTree):
```
pipermerriam: I'm confused.  I think you can drop the base class here and just inherit from `typing.NamedTuple` and get the same effect.

ethpm_cli/install.py
```diff
@@ -47,11 +34,97 @@ def install_package(pkg: Package, config: Config) -> None:
...
+        return self.manifest["package_name"]
+
+    @property
+    def package_version(self) -> str:
+        return self.manifest["version"]
+
+    @property
+    def format_for_display(self) -> str:
+        prefix = "- " * self.depth
+        columns = get_terminal_width()
```
pipermerriam: Some things to think about which might make it simpler to just drop this negotiation of terminal size and let things dump in a less controlled but more predictable format.

1. how does this behave if `len(main_info) + len(hash_info) > columns` which results in a negative `diff`
2. What happens when package names that are very long (maximum 255 chars?)

ethpm_cli/install.py
```diff
@@ -47,11 +34,97 @@ def install_package(pkg: Package, config: Config) -> None:
...
+            )
+        else:
+            children = ""
+        return f"{main_info}{'.' * diff}{hash_info}{children}"
+
+
+def list_installed_packages(config: Config) -> None:
+    installed_packages = [
+        get_installed_package_tree(base_dir)
+        for base_dir in config.ethpm_dir.glob("*/")
```
pipermerriam: Looks like [`Path.iterdir()`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.iterdir) might be the right API to use instead of `glob` here.

ethpm_cli/install.py
```diff
@@ -47,11 +34,97 @@ def install_package(pkg: Package, config: Config) -> None:
...
+    content_hash = ethpm_lock[base_dir.name]["resolved_uri"]
+    dependency_dirs = get_dependency_dirs(base_dir)
+    children = tuple(
+        get_installed_package_tree(dependency_dir, depth + 1)
+        for dependency_dir in dependency_dirs
+    )
+    return InstalledPackageTree(depth, base_dir, manifest, children, content_hash)
+
+
+@to_list
```
pipermerriam: unless there's a reason for the return to be mutable, can this be `to_tuple`?

ethpm_cli/install.py
```diff
@@ -47,11 +34,97 @@ def install_package(pkg: Package, config: Config) -> None:
...
+def get_dependency_dirs(base_dir: Path) -> Iterable[Path]:
+    dep_dir = base_dir / "ethpm_packages"
+    if dep_dir.is_dir():
+        for ddir in dep_dir.glob("*/"):
+            if ddir.is_dir() and ddir.name != "src":
```
pipermerriam: IIUC this correctly, the `src` directory is living alongside the installed dependency packages?  If so I think this needs to be fixed so that dependencies all live in a dedicated directory.

```
# my impression of how it currently is
- ./src/
- ./dep-1/
- ./dep-2/

# how I think it should be
- ./src/
- ./deps/
    - ./dep-1/
    - ./dep-2/
```

This removes any ambiguity about what directories represent dependencies.

Also, assuming my impression of the current structure is correct, what happens if someone makes a package named `src`?

ethpm_cli/install.py
```diff
@@ -47,11 +34,97 @@ def install_package(pkg: Package, config: Config) -> None:
...
+        for dependency_dir in dependency_dirs
+    )
+    return InstalledPackageTree(depth, base_dir, manifest, children, content_hash)
+
+
+@to_list
+def get_dependency_dirs(base_dir: Path) -> Iterable[Path]:
+    dep_dir = base_dir / "ethpm_packages"
+    if dep_dir.is_dir():
+        for ddir in dep_dir.glob("*/"):
```
pipermerriam: same suggestion with `Path.iterdir`

ethpm_cli/install.py
```diff
@@ -47,11 +34,97 @@ def install_package(pkg: Package, config: Config) -> None:
...
+            if ddir.is_dir() and ddir.name != "src":
+                yield ddir
+
+
+def is_package_installed(package_name: str, config: Config) -> bool:
+    return os.path.exists(config.ethpm_dir / package_name)
```
pipermerriam, line 103: Hrm, this raises an interesting connundrum.  This appears to **only** check the alias names.  I think it should probably check both the alias and package name.  Suppose I have the `owned` package installed under the alias `owned-special`.  Then if I checked if `owned` was installed then the CLI would tell me it was not installed.  I think a user would be surprised by this behavior.  

For a package installed with an alias, we could have `uninstall owned` raise a warning that says something like *"found `owned` packages installed under the alias `owned-special`.  You can uninstall this with the comand `uninstall owned-special`"*

ethpm_cli/install.py
```diff
@@ -47,11 +34,97 @@ def install_package(pkg: Package, config: Config) -> None:
...
+
+
+def uninstall_package(package_name: str, config: Config) -> None:
+    if not is_package_installed(package_name, config):
+        raise InstallError(
+            f"Unable to uninstall {package_name} from {config.ethpm_dir}"
```
pipermerriam: This message would confuse me.  I read this as "the package is installed but I can't uninstall it".   Maybe less ambiguous messaging would be: "No package with the name {package_name} found installed under {config.ethpm_dir}"

ethpm_cli/install.py
```diff
@@ -47,11 +34,97 @@ def install_package(pkg: Package, config: Config) -> None:
...
+        )
+
+    tmp_pkg_dir = Path(tempfile.mkdtemp()) / "ethpm_packages"
+    shutil.copytree(config.ethpm_dir, tmp_pkg_dir)
+    shutil.rmtree(tmp_pkg_dir / package_name)
+    uninstall_from_ethpm_lock(package_name, (tmp_pkg_dir / "ethpm.lock"))
```
pipermerriam: `"ethpm.lock"` should probably be a constant.

ethpm_cli/install.py
```diff
@@ -47,11 +34,97 @@ def install_package(pkg: Package, config: Config) -> None:
...
+    return os.path.exists(config.ethpm_dir / package_name)
+
+
+def uninstall_package(package_name: str, config: Config) -> None:
+    if not is_package_installed(package_name, config):
+        raise InstallError(
+            f"Unable to uninstall {package_name} from {config.ethpm_dir}"
+        )
+
+    tmp_pkg_dir = Path(tempfile.mkdtemp()) / "ethpm_packages"
```
pipermerriam: `"ethpm_packages"` should probably be a constant.

ethpm_cli/install.py
```diff
@@ -116,3 +189,9 @@ def update_ethpm_lock(pkg: Package, ethpm_lock: Path) -> None:
     new_pkg_data = pkg.generate_ethpm_lock()
     new_lock = assoc(old_lock, pkg.alias, new_pkg_data)
     ethpm_lock.write_text(f"{json.dumps(new_lock, sort_keys=True, indent=4)}\n")
+
+
+def uninstall_from_ethpm_lock(package_name: str, ethpm_lock: Path) -> None:
+    old_lock = json.loads(ethpm_lock.read_text())
+    new_lock = dissoc(old_lock, package_name)
+    ethpm_lock.write_text(f"{json.dumps(new_lock, sort_keys=True, indent=4)}\n")
```
pipermerriam: This should probably use a tempfile and then move the file into place.  It's prossible for the `json.dumps` to get interrupted before completing the write to disk at which point the lockfile would be corrupt and unreadable.

ethpm_cli/parser.py
```diff
@@ -0,0 +1,59 @@
+import argparse
+from pathlib import Path
+
+
+def get_ethpm_parser() -> argparse.ArgumentParser:
```
pipermerriam, line 5: This is fine, though I think it's also fine to just define the parser at the top level of the module and import it directly rather than rebuilding it each time.

pipermerriam (approved):
I think this is good enough to merge with the commented issues addressed, though writing a few tests that test this using [pexpect](https://pexpect.readthedocs.io/en/stable/) would be a good start towards really knowing this works at the CLI level where its easy to miss errors in how things are all glued together.

ethpm_cli/install.py
njgheorghita: Is this still necessary if `ethpm_lock` is a [tempdir](https://github.com/ethpm/ethpm-cli/pull/17/files#diff-750d4333bcaf89c20dcf7f7469f3d2baR124)?

ethpm_cli/install.py
njgheorghita: Ahh, good catch. the `and ddir.name != "src":` is unnecessary here, we have the structure "how I think it should be" implemented, with `./deps/` as `./ethpm_packages/`.

Good point about the package naming conflict. Both `/src/` and `/ethpm_packages/` are valid package names. I've changed them to `/_src/` and `/_ethpm_packages/`, which are invalid package names so we should never have that conflict, unless you have a better suggestion?

ethpm_cli/install.py
pipermerriam: No, but I don't see anything in this function that requires it to be a temporary directory so I'd be inclined to do it so that it doesn't fail later when someone decides to re-use this function outside of a temporary directory.

PR #23, opened 2019-05-23
Add alias support to uninstall and list

### LINKED ISSUE #20: Check if package is installed under its alias
https://github.com/ethpm/ethpm-cli/pull/17#discussion_r285167725

njgheorghita:
## What was wrong?
Fixes #20 
`ethpm list` now displays package aliases, if package is aliased, alongside package names
`package_name @ alias == version --- (uri)`

#### Cute Animal Picture
image

### DIFF
--- a/ethpm_cli/install.py
+++ b/ethpm_cli/install.py
@@ -23,9 +23,10 @@ logger = logging.getLogger("ethpm_cli.install")
 def install_package(pkg: Package, config: Config) -> None:
     if is_package_installed(pkg.alias, config):
         raise InstallError(
-            "Installation conflict: A directory or file already exists at the install location "
-            f"for the package '{pkg.manifest['package_name']}' aliased to '{pkg.alias}' on the "
-            f"filesystem at {config.ethpm_dir / pkg.alias}."
+            f"Installation conflict: Package: '{pkg.manifest['package_name']}' "
+            f"aliased to '{pkg.alias}' already installed on the filesystem at "
+            f"{config.ethpm_dir / pkg.alias}. Try installing this package with "
+            "a different alias."
         )
 
     # Create temporary package directory
@@ -57,7 +58,11 @@ class InstalledPackageTree(NamedTuple):
     @property
     def format_for_display(self) -> str:
         prefix = "- " * self.depth
-        main_info = f"{prefix}{self.package_name}=={self.package_version}"
+        if self.path.name != self.package_name:
+            alias = f" @ {self.path.name}"
+        else:
+            alias = ""
+        main_info = f"{prefix}{self.package_name}{alias}=={self.package_version}"
         hash_info = f"({self.content_hash})"
         if self.children:
             children = "\n" + "\n".join(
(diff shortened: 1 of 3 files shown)


ethpm_cli/install.py
```diff
@@ -57,7 +58,10 @@ def package_version(self) -> str:
     @property
     def format_for_display(self) -> str:
         prefix = "- " * self.depth
-        main_info = f"{prefix}{self.package_name}=={self.package_version}"
+        if self.path.name != self.package_name:
```
pipermerriam, line 61: Doesn't have to be addressed now but probably worth opening a ticket for this.  This mechanism, while totally functional, is very indirect, and thus, likely to become error prone at some point.  It would be ideal if we sourced the alias from a "reliable" source that was intended to be a store of information.

ethpm_cli/install.py
```diff
@@ -103,20 +107,42 @@ def is_package_installed(package_name: str, config: Config) -> bool:
...
+        uninstall_from_ethpm_lock(package_name, (tmp_pkg_dir / LOCKFILE_NAME))
+
+        shutil.rmtree(config.ethpm_dir)
+        tmp_pkg_dir.replace(config.ethpm_dir)
+        return
+
+    aliases = get_package_aliases(package_name, config)
+    if aliases:
+        raise InstallError(
+            f"Found {package_name} installed under the alias(es): {aliases}. "
```
pipermerriam, line 139: :+1: , :heart: great error messaging.

tests/cli/test_ethpm.py
```diff
@@ -24,6 +24,17 @@ def test_ethpm_list(test_assets_dir):
     )
 
 
+def test_ethpm_list_with_aliased_package(test_assets_dir):
+    ethpm_dir = test_assets_dir / "owned" / "ipfs_uri_alias" / "_ethpm_packages"
+    child = pexpect.spawn(f"ethpm list --ethpm-dir {ethpm_dir}")
+    child.expect("EthPM CLI v0.1.0a0\r\n")
+    child.expect("\r\n")
+    child.expect(
+        "owned @ owned-alias==1.0.0 --- "
```
pipermerriam, line 33: how are you liking testing with `pexpect`  I admit I've not used it myself so I'm curious.

Does it support any kind of easy regex/pattern matching?

tests/cli/test_ethpm.py
```diff
@@ -53,3 +64,29 @@ def test_ethpm_uninstall(tmp_path, test_assets_dir):
     assert check_dir_trees_equal(
         ethpm_dir, test_assets_dir / "wallet" / "ipfs_uri" / "_ethpm_packages"
     )
+
+
+def test_ethpm_uninstall_nonexistent_package(tmp_path):
+    ethpm_dir = tmp_path / "_ethpm_packages"
```
pipermerriam, line 80: I'm not sure if you're doing this, but to prevent any accidental mucking up of your local working directory I recommend an `autouse=True` fixture in your soot conftest like this:

```python
@pytest.fixture(autouse=True)
def _clean_current_working_directory(monkeypatch):
    with tempfile.tempdir() as temp_dir:
        monkeypatch.chdir(temp_dir)
        yield
```

pipermerriam (approved):

tests/cli/test_ethpm.py
njgheorghita: Yup, it supports regex and I believe it's just using the `re` module [match object](https://docs.python.org/3/library/re.html#match-objects) - though I've been using it mainly to just match strings. 

Debugging was a little tedious at first, but after getting more familiar with the api it's been pretty smooth.

PR #26, opened 2019-05-24
Write some utils to help manage solidity files

njgheorghita:
## What was wrong?
A couple utils i've been using locally to quickly & easily generate the most basic manifest possible from solc output. Not sure if these are better suited here or in `py-ethpm` though?

#### Cute Animal Picture
image

### DIFF
--- a/ethpm_cli/__init__.py
+++ b/ethpm_cli/__init__.py
@@ -1,5 +1,7 @@
 from pathlib import Path
 
-CLI_ASSETS_DIR: Path = Path(__file__).parent / "assets"
+ETHPM_CLI_DIR = Path(__file__).parent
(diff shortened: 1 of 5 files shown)


pipermerriam:
These are hard to evaluate without context.  Maybe some docs for how you would use these tools to do something like...

```
$ ethpm <do-thing-that-makes-json-configuration> | solidity <do-the-compilation> | ethpm <get-the-compiled-results-and-make-package> --outfile super-package.v1.json
```

Basically, the python functions are probably useful to you, but what does the full user-friendly-enough-pipeline look like.

njgheorghita:
closing this since #33 takes these utils and integrates them into a user workflow

PR #27, opened 2019-06-11
Install packages from Etherscan verified contracts

njgheorghita:
## What was wrong?
Add feature to install etherscan verified contracts as ethpm packages

`ethpm install --etherscan {contract_addr} --package-name pkg --version 1.0.0`

Refactored some tests into their own files

#### Cute Animal Picture
image


ethpm_cli/parser.py
```diff
@@ -76,23 +76,49 @@ def scrape_action(args: argparse.Namespace) -> None:
...
 
 install_parser = ethpm_parser.add_parser("install", help="Install uri")
 install_parser.add_argument(
-    "uri",
+    "--uri",
     action="store",
     type=str,
     help="IPFS / Github / Registry URI of package you want to install.",
 )
+install_parser.add_argument(
```
pipermerriam: Take a look at `add_mutually_exclusive_group` since it appears that `--uri` and `--etherscan` should not be provided at the same time.

ethpm_cli/parser.py
```diff
@@ -76,23 +76,49 @@ def scrape_action(args: argparse.Namespace) -> None:
...
+    "--etherscan",
+    dest="etherscan",
+    action="store",
+    type=str,
+    help="Verified mainnet contract address.",
```
pipermerriam: What do you think about defining a URI scheme for etherscan packages?  That way we don't need a special argument and it could just be one of the backends.

pipermerriam (commented):
Probably time to write some documentation, even if it's just a page of doctest examples of how to do common things..  API docs can come later.

ethpm_cli/parser.py
njgheorghita: @pipermerriam I'm on board for that, makes it easy to support multiple chains. What do you think 
about something like ...`etherscan://CONTRACT_ADDR/CHAIN_ID`

njgheorghita: maybe even the trailing `CHAIN_ID` is optional and defaults to the mainnet?

pipermerriam: My ideas:

- `etherscan://ADDR@CHAIN_ID`
- `etherscan://ADDR?chain_id=CHAIN_ID`
- `etherscan://ADDR.mainnet` 
- `etherscan://ADDR/CHAIN_ID` (your suggestion)

It's easier to relax constraints than add them so lets go with your idea but make it required for now.

njgheorghita:
@pipermerriam having some trouble getting these tests working in circleci env, but this pr's ready for review

ethpm_cli/_utils/filesystem.py
```diff
@@ -58,3 +61,7 @@ def atomic_replace(path: Path) -> Generator[IO[Any], None, None]:
         with tmp_file_path.open(mode="w+") as tmpfile:
             yield tmpfile
         shutil.copyfile(tmp_file_path, path)
+
+
+def is_package_installed(package_name: str, config: "Config") -> bool:
+    return os.path.exists(config.ethpm_dir / package_name)
```
pipermerriam: What other files or directory names might be present that this would return a false positive for?  Are there further checks  you could do about the path that would improve accuracy?  Ensure it is a directory?  Ensure it has certain files present in it?

ethpm_cli/validation.py
```diff
@@ -50,25 +62,63 @@ def validate_solc_output(project_dir: Path) -> None:
...
         validate_ethpm_dir(args.ethpm_dir)
 
 
-def validate_target_uri(uri: URI) -> None:
-    if not is_supported_content_addressed_uri(uri) and not is_valid_registry_uri(uri):
+def validate_package_is_installed(contract_type_id: str, config: "Config") -> None:
+    package, contract_type = contract_type_id.split(":")
+    validate_package_name(package)
+    if not is_package_installed(package, config):
+        raise InstallError(f"{package} is not installed.")
```
pipermerriam: Better error message might include the names of all of the packages that are installed.

ethpm_cli/_utils/etherscan.py
```diff
@@ -0,0 +1,37 @@
+from typing import Any
+from urllib import parse
+
+from eth_utils import is_checksum_address
+
+ETHERSCAN_SUPPORTED_CHAIN_IDS = {
+    "1": "",
```
pipermerriam: might be worth just commenting that this is `'mainnet'`

pipermerriam (approved):

PR #28, opened 2019-06-15
Write documentation to disk if present in manifest

njgheorghita:
## What was wrong?
Now `ethpm install` will write documentation to disk under `documentation.md` if an ipfs uri is present in manifest metadata

#### Cute Animal Picture
image


pipermerriam (approved):
Do any of the tests cover this?  I see the documentation file is present in many of the fixture assets but it isn't clear whether that means there is test coverage.  Might be good to have at least one unit test that calls this method directly and verifies that expected file is written.

njgheorghita:
Yup, there are about 5 tests that already cover this.

PR #29, opened 2019-06-16
Remove duplicate validation

njgheorghita:
## What was wrong?
URI backends from `ethpm` [already](https://github.com/ethpm/py-ethpm/blob/master/ethpm/backends/ipfs.py#L62) perform [validation](https://github.com/ethpm/py-ethpm/blob/master/ethpm/backends/http.py#L31) of resolved contents and uri content hashes.

#### Cute Animal Picture
image


njgheorghita:
merging since this is a minor change

PR #34, opened 2019-07-11
Write initial authorization utils

njgheorghita:
## What was wrong?
Creates a command to use a password-protected keyfile for auth purposes.

Set environment variables via.
```
ethpm auth <path_to_keyfile> <password>
```

#### Cute Animal Picture
image

### DIFF
--- a/ethpm_cli/auth.py
+++ b/ethpm_cli/auth.py
@@ -2,10 +2,18 @@ import os
 from typing import Optional
 
 from eth_account import Account
+import eth_keyfile
 from eth_typing import Address
+from eth_utils import to_bytes
 
-from ethpm_cli.constants import PRIVATE_KEY_ENV_VAR
+from ethpm_cli.constants import KEYFILE_PASSWORD, KEYFILE_PATH
 from ethpm_cli.exceptions import ValidationError
+
+
+def set_auth(keyfile_path: str, password: str) -> None:
+    validate_keyfile_path_and_password(keyfile_path, to_bytes(text=password))
+    os.environ[KEYFILE_PATH] = keyfile_path
+    os.environ[KEYFILE_PASSWORD] = password
 
 
 def get_authorized_address() -> Address:
@@ -14,16 +22,19 @@ def get_authorized_address() -> Address:
 
 
 def get_authorized_private_key() -> Optional[str]:
-    private_key = os.getenv(PRIVATE_KEY_ENV_VAR)
-    if private_key:
-        validate_private_key(private_key)
-        return private_key
-    return None
-
-
-def validate_private_key(key: str) -> None:
-    if len(key) != 64:
+    keyfile_path = os.getenv(KEYFILE_PATH)
+    password = to_bytes(text=os.getenv(KEYFILE_PASSWORD))
+    if not keyfile_path or not password:
         raise ValidationError(
-            f"Private key ({key}) stored under environment variable: "
-            f"{PRIVATE_KEY_ENV_VAR} is not a valid private key."
+            "Must set password-protected keyfile via `ethpm auth` command."
         )
+    validate_keyfile_path_and_password(keyfile_path, password)
+    private_key = eth_keyfile.extract_key_from_keyfile(keyfile_path, password)
+    return private_key
+
+
+def validate_keyfile_path_and_password(keyfile_path: str, password: str) -> None:
+    try:
+        eth_keyfile.extract_key_from_keyfile(keyfile_path, password)
+    except ValueError:
+        raise ValidationError(f"Invalid keyfile password.")
--- a/ethpm_cli/auth.py
+++ b/ethpm_cli/auth.py
@@ -1,6 +1,6 @@
 from pathlib import Path
 import tempfile
-from typing import Any, Dict, Optional
+from typing import Any, Dict
 
 import eth_keyfile
 from eth_typing import Address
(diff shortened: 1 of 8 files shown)


pipermerriam:
Do you know https://github.com/ethereum/eth-keyfile exists?  That would allow us to use encrypted keyfiles which is probably approximately the same complexity and should be much more secure.

pipermerriam (commented):
seeing how simple this is I think I'm even more inclined to push for a keyfile based approach.  Should be almost exactly as simple but with a single *prompt* for the password to the keyfile which can go in a later PR, for now you could just write the APIs as if you already have the password.

njgheorghita:
@pipermerriam Ping for another review - this pr's been updated to work with `eth-keyfile`

ethpm_cli/auth.py
```diff
@@ -0,0 +1,40 @@
...
+from eth_utils import to_bytes
+
+from ethpm_cli.constants import KEYFILE_PASSWORD, KEYFILE_PATH
+from ethpm_cli.exceptions import ValidationError
+
+
+def set_auth(keyfile_path: str, password: str) -> None:
+    validate_keyfile_path_and_password(keyfile_path, to_bytes(text=password))
+    os.environ[KEYFILE_PATH] = keyfile_path
+    os.environ[KEYFILE_PASSWORD] = password
```
pipermerriam: This gives me pause.  Is there a reason you need to set these in the environment.  IIUC the motivation for this is so that you can set these up using a CLI command and then use them later?

I think this might be a problematic approach due to it exposing the key in the keyfile in a manner that is *effectively* storing it in plain text.  My suggestion for how we might do this in a manner that preserves security but still gives good UX.

1. Split this into two mechanisms.
    - A mechanism for recording the location of the keyfile that a user wishes to use for future commands.  I'd recommend using the filesystem and maybe a config file located somewhere like [`$XDG_CONFIG_HOME/ethpm_cli/`](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)
        - maybe a separate mechanism that allows *importing* a keyfile which actually copies the file into `XDG_CONFIG_HOME/ethpm_cli/keystore` for easy future use?
    - A mechanism for the **user** to provide the password via multiple mechanisms.
        - CLI prompt where they type it in
        - CLI flag where they provide it in plaintext
        - environment variable where they set it in `os.environ`

pipermerriam (commented):
If my suggestion doesn't make sense lets have a call to discuss.

ethpm_cli/auth.py
```diff
@@ -0,0 +1,62 @@
...
+from ethpm_cli._utils.xdg import get_xdg_ethpmcli_root
+
+
+def import_keyfile(keyfile_path: Path):
+    validate_keyfile(keyfile_path)
+    ethpm_xdg_root = get_xdg_ethpmcli_root()
+    ethpm_cli_keyfile_path = ethpm_xdg_root / KEYFILE_PATH
+    tmp_keyfile = Path(tempfile.NamedTemporaryFile().name)
+    tmp_keyfile.write_text(keyfile_path.read_text())
+    tmp_keyfile.replace(ethpm_cli_keyfile_path)
```
pipermerriam, line 20: nice way to do atomic write :+1:

ethpm_cli/auth.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def get_keyfile_path():
+    ethpm_xdg_root = get_xdg_ethpmcli_root()
+    keyfile_path = ethpm_xdg_root / KEYFILE_PATH
+    if not keyfile_path.exists():
+        raise Exception
```
pipermerriam: maybe `FileNotFoundError`?

ethpm_cli/auth.py
```diff
@@ -0,0 +1,62 @@
...
+
+def get_authorized_address() -> Address:
+    """
+    Returns the address associated with stored keyfile.
+    """
+    keyfile = get_keyfile_data()
+    return keyfile["address"]
+
+
+def get_authorized_private_key(password: str) -> Optional[str]:
```
pipermerriam: I'm confused by the `Optional` here.  Does `eth_keyfile.extract_key_from_keyfile` have a `None` return path?

njgheorghita:
@pipermerriam Makes sense. I've updated the strategy to look like 
```
ethpm auth
```
will display the address currently associated with the stored keyfile - or a message saying no keyfile is available

```
ethpm auth --keyfile-path <path>
```
will set your keyfile (aka. copies the keyfile into ethpm xdg root). In terms of the password, it makes sense to me (from a security perspective) to simply require it as a flag for all relevant cli commands. No commands so far need a password, so I'll make that update in a following pr.

Let me know how that sounds to you

tests/core/test_auth.py
```diff
@@ -0,0 +1,48 @@
...
+PASSWORD = b"password"
+KEYFILE_JSON = eth_keyfile.create_keyfile_json(PRIVATE_KEY, PASSWORD)
+
+
+@pytest.fixture
+def keyfile(monkeypatch, tmpdir):
+    monkeypatch.chdir(tmpdir)
+    ethpmcli_dir = Path(tmpdir / "ethpmcli_dir")
+    ethpmcli_dir.mkdir()
+    monkeypatch.setenv("XDG_ETHPMCLI_ROOT", str(ethpmcli_dir))
```
pipermerriam: This is probably something that you want to be in the root `conftest.py` file and to flag it with `autouse=True` **AND** to have a test that sanity checks that none of the utils that fetch this directory return the *real* directory so that you have strong confidence that you aren't accidentally modifying the actual user director during test runs.

pipermerriam (approved):
Looks good.

PR #35, opened 2019-07-15
Registry admin commands

njgheorghita:
## What was wrong?
Commands were needed to help users manage their ethpm registries.

```
ethpm registry list
```
- Lists all of the stored registries & aliases available to the user

```
ethpm registry add <uri> --alias <alias>
```
- Adds a registry to the local registry store

```
ethpm registry activate <uri_or_alias>
```
- "Activates" a registry if found in registry store. "Activated" registries are simply the default registry used for publishing new packages. To do so, user must also use `ethpm auth` to manage private keys if registry has auth restrictions.

#### Cute Animal Picture
image


njgheorghita:
@pipermerriam ready for review but tests are blocked until https://github.com/ethereum/web3.py/pull/1398/ gets merged

njgheorghita:
ping @pipermerriam for :eyes:. Now with new web3 version and passing tests

ethpm_cli/registry.py
```diff
@@ -0,0 +1,156 @@
...
+class StoredRegistry(NamedTuple):
+    uri: URI
+    active: bool = False
+    alias: Optional[str] = None
+    ens: Optional[str] = None
+
+    @property
+    def format_for_display(self) -> str:
+        activated = "(active)" if self.active else ""
+        return f"{self.uri} --- {self.alias} {activated}"
```
pipermerriam: You might include the leading whitespace in the `activated` variable so that there isn't a dangling space when it isn't activated.

```python
activated = " (active)" if self.active else ""
return f"{self.uri} --- {self.alias}{activated}"
```

ethpm_cli/registry.py
```diff
@@ -0,0 +1,156 @@
...
+def generate_registry_store_data(
+    registry_uri: URI, alias: str, activate: bool = False
+) -> Iterable[Tuple[str, Any]]:
+    parsed_uri = parse_registry_uri(registry_uri)
+    # todo: support ens in registry uri
+    ens = None
+    yield "ens", ens
+    yield "address", parsed_uri.address
+    yield "alias", alias
+    yield "active", activate
```
pipermerriam: Since there's no logic in this function to exclude any keys you can just construct and return the dictionary without using the `to_dict` approach.  probably a little simpler to deal with.

pipermerriam (approved):

PR #39, opened 2019-07-24
Move xdg directory setup to Config

njgheorghita:
## What was wrong?
Extended the `Config` class to handle initializing / validating the ethpm xdg directory.
Also extended the `Config` class to initialize a `w3` instance for use based on an optional `chain-id` cli arg.

#### Cute Animal Picture
image


ethpm_cli/config.py
```diff
@@ -29,3 +38,43 @@ def __init__(self, args: Namespace) -> None:
...
                 self.ethpm_dir.mkdir()
         validate_ethpm_dir(self.ethpm_dir)
+
+        # Setup xdg ethpm dir
+        xdg_ethpm_dir = get_xdg_ethpmcli_root()
+        w3 = Web3(load_provider_from_uri(INFURA_HTTP_URI))
+        if xdg_ethpm_dir.is_dir():
+            validate_xdg_ethpm_dir(xdg_ethpm_dir, w3)
+        else:
+            initialize_xdg_ethpm_dir(xdg_ethpm_dir, w3)
```
pipermerriam: I don't know if it's possible but if you can make `initialize_xdg_ethpm_dir` idempotent then you can simplify this to just be something that is unconditionally called at this stage and include the validation in the initialization logic.

ethpm_cli/config.py
```diff
@@ -29,3 +38,43 @@ def __init__(self, args: Namespace) -> None:
...
+    write_updated_chain_data(chain_data_path, init_json)
+    ethpm_keyfile = ethpm_dir / KEYFILE_PATH
+    ethpm_keyfile.touch()
+
+
+def write_updated_chain_data(
+    chain_data_path: Path, updated_data: Dict[str, Any]
+) -> None:
+    tmp_pkg_dir = Path(tempfile.mkdtemp())
+    tmp_data = tmp_pkg_dir / "chain_data.json"
```
pipermerriam: I realize you could encapsulate this into a utility like this.

```python
with atomic_replace(Path("some/path.txt")) as file:
    file.write('some text')
    # here nothing has actually been done to the original file
# here the context manager has replaced the original file
```

Seems like a simple and nice utility to do this concisely

njgheorghita:
ping @pipermerriam for another :eyes:

ethpm_cli/parser.py
```diff
@@ -194,11 +206,5 @@ def list_action(args: argparse.Namespace) -> None:
 list_parser = ethpm_parser.add_parser(
     "list", help="List all installed packages in your ethPM directory."
 )
-list_parser.add_argument(
-    "--ethpm-dir",
```
pipermerriam, line 198: Curious about your choice to include this in all of these parsers rather than include it a single time in the main parser?

tests/core/_utils/test_filesystem.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def test_atomic_replace_is_atomic(original):
+    try:
+        with atomic_replace(original) as original_file:
+            original_file.write(json.dumps({"three": "four"}))
+            raise Exception
+    except Exception:
+        pass
+    assert "three" not in json.loads(original.read_text()).keys()
```
pipermerriam: might be worth also asserting the original value is still present.

tests/core/_utils/test_filesystem.py
```diff
@@ -0,0 +1,28 @@
...
+
+import pytest
+
+from ethpm_cli._utils.filesystem import atomic_replace
+
+
+@pytest.fixture
+def original(tmp_path):
+    original = tmp_path / "original.json"
+    original.write_text(json.dumps({"one": "two"}))
```
pipermerriam: you could simplify these tests by dropping the json loading/dumping and just writing/reading basic strings to/from the files and asserting they are present/missing

pipermerriam (approved):
This is a pretty large and scattered diff.  Found it kind of hard to read so my approval here isn't signaling a thorough review but rather a scan for general structure.

ethpm_cli/parser.py
njgheorghita: It was mostly an aesthetic decision. Preferring something like
```
ethpm install <uri> --package-name xxx --version 1 --ethpm-dir /path
```
over
```
ethpm --ethpm-dir /path install <uri> --package-name xxx --version 1
```
I prefer putting the `ethpm-dir` arg at the end of the command, but maybe it's relatively insignificant? And I should just include the arg on the main parser, and simplify the code?

njgheorghita:
@pipermerriam Yeah, sorry about the diff. This pr got a little more scatterd than I would have liked, but I'll try to simplify future prs.

ethpm_cli/parser.py
pipermerriam: That makes sense.  I think this is the API you're looking for to remove the repitition.

https://docs.python.org/3/library/argparse.html#parents

pipermerriam: Ok, maybe not since that seems more about the main parser and not the subparsers..  not sure, carry on.

PR #40, opened 2019-07-31
Update overview docs

njgheorghita:
## What was wrong?

More docs.

#### Cute Animal Picture

image

### DIFF
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -1,0 +1,67 @@
+Commands
+========
+
+A command-line tool to help manage EthPM packages and registries.
+
+
+.. warning::
+
+   ``EthPM CLI`` is currently in public Alpha:
+
+   - It is expected to have bugs and is not meant to be used in production 
+   - Things may be ridiculously slow or not work at all
+
+
+ethpm install
+-------------
+
+Install an ethPM package to a local ``_ethpm_packages`` directory.
+
+.. argparse::
+   :ref: ethpm_cli.parser.parser
+   :prog: ethpm
+   :path: install
+
+
+ethpm list
+----------
+
+List all installed ethPM packages in a local ``_ethpm_packages`` directory.
+
+.. argparse::
+   :ref: ethpm_cli.parser.parser
+   :prog: ethpm
+   :path: list
+
+
+ethpm uninstall
+---------------
+
+Uninstall an ethPM package from a local ``_ethpm_packages`` directory.
+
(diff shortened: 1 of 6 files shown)


njgheorghita:
Merging since I'd like to see if this pr fixes my readthedocs building bug. But if @pipermerriam you have any posthumous comments, that would be 👍

docs/commands.rst
```diff
@@ -0,0 +1,67 @@
...
+.. argparse::
+   :ref: ethpm_cli.parser.parser
+   :prog: ethpm
+   :path: uninstall
+
+
+ethpm auth
+----------
+
+Link a keyfile to authorize on-chain transactions (i.e. deploying a registry / releasing a package). To generate a keyfile, use `eth-keyfile <https://github.com/ethereum/eth-keyfile>`_.
```
pipermerriam, line 51: Might be worth providing a full example python script to make this extra easy.

docs/disk.rst
```diff
@@ -1,7 +1,10 @@
 Disk Format
 ===========
 
-ethPM cli writes package assets to your disk using the following format.
+_ethpm_packages/
+----------------
+
+ethPM cli writes ethPM package assets to your disk using the following format. You can have multiple different ``_ethpm_packages/`` local directories. The ``--ethpm-dir`` flag is available on most ``ethpm-cli`` commands to target a specific ``_ethpm_packages/`` dir. You can also set the environment variable ``ETHPM_CLI_PACKAGES_DIR`` to your target ``_ethpm_packages/`` directory. Otherwise, the cli will automatically lookup any ``_ethpm_packages/`` directory available under the current working dirctory.
```
pipermerriam, line 7: This is probably better laid out so that it's clear what order of precedence these will go in.  Something like this.

- by default it will write files to `./_ethpm_packages/` in the current working directory.
- If `--ethpm-dir` is provided then the CLI will use this.
- If the environment variable `ETHPM_CLI_PACKAGES_DIR` is set the CLI will use this if one is not specified using the `--ethpm-dir` flag

pipermerriam (commented):
:+1: looks like the docs may not be rendering whatever fancy argparse stuff you're trying to get working.

docs/commands.rst
njgheorghita: Thoughts on also providing a cli command to do this?
```
ethpm auth generate --private-key xxx --password yyy
```

PR #50, opened 2019-08-02
Add deployments to manifest wizard

njgheorghita:
## What was wrong?
Extended the manifest wizard to support adding a single or multiple deployments.

#### Cute Animal Picture

image


ethpm_cli/manifest.py
```diff
@@ -99,21 +106,103 @@ def gen_sources(
...
+    while chain_id not in SUPPORTED_CHAIN_IDS.keys():
+        cli_logger.info(f"{chain_id} is not a supported chain id. ")
+        chain_id = to_int(text=input(question))
+    return chain_id
+
+
+def get_deployment_address() -> str:
+    question = "What is the address of your deployment? "
+    address = input(question)
+    while not is_address(address):
```
pipermerriam: probably want to do `is_checksum_address` here and enforce checksums on user inputs.

ethpm_cli/manifest.py
```diff
@@ -68,7 +74,7 @@ def generate_custom_manifest(project_dir: Path) -> None:
 
 def gen_validate_manifest() -> Optional[Callable[..., Manifest]]:
     flag = parse_bool_flag(
-        "Would you like to validate your manifest against the json schema? (recommended) (y/n) "
+        "Would you like to validate your manifest against the json schema? (recommended)"
```
pipermerriam, line 77: Maybe extend `parse_bool_flag` to support a `default` so that the user can just pres `Enter` without actually typing in a `y` or `n`.  My intuition says this would reduce mistakes where a user is repeatedly typing in `y` to get through the prompts and says `y` to something they don't intend to.

pipermerriam (approved):

ethpm_cli/manifest.py
njgheorghita: Yeah, I could see that being helpful but maybe down the road after the wizard is a little more fleshed out. It's just not clear to me whether or not it should always default to `y`? Like for adding links to the manifest, I'm not sure we should default to making the user do that and every other possible field? Or should the default change depending on the specific question?

pipermerriam: The default would change depending on the specific question.

PR #57, opened 2019-08-14
Generate contract types to automatically include sources

njgheorghita:
## What was wrong?
Previously, all contract types and sources available were automatically added. This is not useful when dealing with large libraries of contracts. Also, the burden is on the user to make sure that the required sources are included for each specified contract type. 

Now, the user only has to specify "all available contract types" or "a comma separated list of desired contract types" to include in the manifest. The wizard will automatically include all necessary source files for all included contract types. 

#### Cute Animal Picture

image

### DIFF
--- a/ethpm_cli/_utils/solc.py
+++ b/ethpm_cli/_utils/solc.py
@@ -1,12 +1,13 @@
 import json
 from pathlib import Path
-from typing import Any, Callable, Dict, Iterable
+from typing import Any, Callable, Dict, Iterable, Tuple
 
 from eth_typing import Manifest
 from eth_utils import to_tuple
 from eth_utils.toolz import assoc
 from ethpm.tools import builder as b
 
+from ethpm_cli._utils.ipfs import get_ipfs_backend
 from ethpm_cli._utils.logger import cli_logger
 from ethpm_cli.constants import SOLC_INPUT, SOLC_OUTPUT
 
@@ -50,11 +51,21 @@ def generate_solc_input(contracts_dir: Path) -> None:
     )
 
 
-def build_sources(
+def build_inline_sources(
     contract_types: Iterable[str], solc_output: Dict[str, Any], contracts_dir: Path
 ) -> Iterable[Callable[..., Manifest]]:
     return (
         b.inline_source(ctype, solc_output, contracts_dir) for ctype in contract_types
+    )
+
+
+def build_pinned_sources(
+    contract_types: Iterable[str], solc_output: Dict[str, Any], contracts_dir: Path
+) -> Iterable[Callable[..., Manifest]]:
+    ipfs_backend = get_ipfs_backend()
+    return (
+        b.pin_source(ctype, solc_output, ipfs_backend, contracts_dir)
+        for ctype in contract_types
     )
 
 
@@ -69,7 +80,7 @@ def create_basic_manifest_from_solc_output(
 ) -> Manifest:
     solc_output = json.loads((project_dir / SOLC_OUTPUT).read_text())["contracts"]
     contract_types = get_contract_types(solc_output)
-    built_sources = build_sources(
+    built_sources = build_inline_sources(
         contract_types, solc_output, project_dir / "contracts"
     )
     built_types = build_contract_types(contract_types, solc_output)
@@ -89,3 +100,14 @@ def get_contract_types(solc_output: Dict[str, Any]) -> Iterable[str]:
     for source in solc_output:
         for ctype in solc_output[source].keys():
             yield ctype
+
+
+@to_tuple
+def get_contract_types_and_sources(
+    solc_output: Dict[str, Any]
+) -> Iterable[Tuple[str, Iterable[Path]]]:
+    for source in solc_output:
+        for ctype, data in solc_output[source].items():
+            metadata = json.loads(data["metadata"])
+            sources = tuple(Path(src) for src in metadata["sources"].keys())
+            yield ctype, sources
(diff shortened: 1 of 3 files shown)


ethpm_cli/_utils/solc.py
```diff
@@ -89,3 +100,14 @@ def get_contract_types(solc_output: Dict[str, Any]) -> Iterable[str]:
     for source in solc_output:
         for ctype in solc_output[source].keys():
             yield ctype
+
+
+@to_tuple
+def get_contract_types_and_sources(
+    solc_output: Dict[str, Any]
+) -> Iterable[Tuple[str, Any]]:
```
pipermerriam: Seems like you can be more specific about the return type here.  Looks like it's going to be an iterable of some kind of strings representing file paths?

ethpm_cli/manifest.py
```diff
@@ -81,45 +86,63 @@ def gen_validate_manifest() -> Optional[Callable[..., Manifest]]:
...
+    else:
+        while True:
+            raw_included_ctypes = input(
+                "Please list the contract types you would like to include, separated by commas: "
+            )
+            target_contract_types = [
+                ct.strip(" ") for ct in raw_included_ctypes.split(",")
+            ]
+            if any(ct not in all_contract_types for ct in target_contract_types):
+                cli_logger.info(f"Invalid contract type selected. Please try again.")
```
pipermerriam: Good to tell them which one was invalid.

ethpm_cli/manifest.py
```diff
@@ -81,45 +86,63 @@ def gen_validate_manifest() -> Optional[Callable[..., Manifest]]:
...
-        f"{len(available_sources)} sources available.\n\n"
-        f"{pretty}. \n\n"
-        "Would you like to include all available sources?"
-    )
-    if not flag:
-        raise Exception("sorry, we dont support specific source selection yet.")
     inline_source_flag = parse_bool_flag(
-        "Would you like to automatically inline all sources?"
+        "Would you like to inline source files? If not, sources will "
+        "be automatically pinned to IPFS."
```
pipermerriam, line 128: Random thought.

When we pin an IPFS asset, we could also write it to the ethpm IPFS cache.  One *neat* option that would arise from this would be the ability to re-pin assets that get lost.  You could run a command that iterates over the items in the cache and re-pins any that can't be found on the IPFS network.

ethpm_cli/manifest.py
```diff
@@ -81,45 +86,63 @@ def gen_validate_manifest() -> Optional[Callable[..., Manifest]]:
...
+    )
+    target_sources = set(
```
pipermerriam, line 130: you can use a set comprehension here `{thing for thing in things}`

ethpm_cli/manifest.py
```diff
@@ -81,45 +86,63 @@ def gen_validate_manifest() -> Optional[Callable[..., Manifest]]:
...
+        flatten([ct[1] for ct in ctypes_and_sources if ct[0] in target_contract_types])
     )
-    if not inline_source_flag:
-        raise Exception("Sorry, we dont support pinning sources yet.")
-    # todo: validate a source for every contract type
-    return build_sources(contract_types, solc_output, contracts_dir)
+    target_source_names = (Path(src).stem for src in target_sources)
```
pipermerriam: this will be a generator.  Is `build_contract_types` guaranteed to work properly with generators?  If there isn't a test for that I'd advocate fully evaluating this to be a `tuple` or something.

ethpm_cli/manifest.py
```diff
@@ -81,45 +86,63 @@ def gen_validate_manifest() -> Optional[Callable[..., Manifest]]:
...
-    )
-    if not flag:
-        raise Exception("sorry, we dont support specific source selection yet.")
     inline_source_flag = parse_bool_flag(
-        "Would you like to automatically inline all sources?"
+        "Would you like to inline source files? If not, sources will "
+        "be automatically pinned to IPFS."
+    )
+    target_sources = set(
+        flatten([ct[1] for ct in ctypes_and_sources if ct[0] in target_contract_types])
```
pipermerriam: Maybe rather than indexing into `ct` you could do:

```
flatten(source_path for ctype, source_path in ctypes_and_sources if ctype in target_contract_types)
```

ethpm_cli/_utils/solc.py
```diff
@@ -89,3 +100,14 @@ def get_contract_types(solc_output: Dict[str, Any]) -> Iterable[str]:
...
+    for source in solc_output:
+        for ctype, data in solc_output[source].items():
+            metadata = json.loads(data["metadata"])
+            sources = metadata["sources"].keys()
+            yield (ctype, sources)
```
pipermerriam: You might consider normalizing the `sources` to be `Path` objects at this stage.  Saves you some `Path` casting elsewhere.

ethpm_cli/manifest.py
```diff
@@ -81,45 +86,63 @@ def gen_validate_manifest() -> Optional[Callable[..., Manifest]]:
...
+    return ((*generated_contract_types), (*generated_sources))
+
+
+@to_list
+def format_contract_types_and_sources_for_display(
+    ctypes_and_sources: Tuple[str]
+) -> Iterable[str]:
+    for ctype in ctypes_and_sources:
+        yield f"{ctype[0]}\n"
+        for src in ctype[1]:
```
pipermerriam: I think you *likely* have a pending test failure in CI or something due to these values not having their order guaranteed.  To fix this you should probably add a `sorted` call somewhere in here to ensure that the data is output in a deterministic order.

pipermerriam (approved):

PR #59, opened 2019-08-20
Remove unnecessary ethpmdir flag from release command

njgheorghita:
## What was wrong?
Unnecessary flag

#### Cute Animal Picture
image


njgheorghita:
merging since minor

Issue #31, opened 2019-07-02
Installation strategies


njgheorghita:
Make cli tool available for installation via both Pypi and Homebrew.

cgewecke:
@njgheorghita Do you have any thoughts about also publishing a docker image? Solidity, Vyper and Geth all do this and it's made them pretty straightforward to consume in the JS build pipeline tools.

Happy to look into this and open a PR here if it's something you think worth doing...

njgheorghita:
@cgewecke Yup, that's a great thought, thanks. A pr would be great, but you might want to wait until some kind of stable alpha is released? Which should hopefully be in the very near future.

cgewecke:
@njgheorghita Yes sounds good!

njgheorghita:
Hijacking this issue to update that we should also have a distribution via apt for linux users.

PR #61, opened 2019-08-21
Add installation docs

njgheorghita:
## What was wrong?
Installation docs

#### Cute Animal Picture
image

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -1,9 +1,42 @@
 Installation
 ------------
 
-Current ethPM CLI is only available via Pypi.
-
 Pypi
 ~~~~
 
-``pip install ethpm-cli``
+- Create your virtual environment
+- ``pip install ethpm-cli``
+
+
+Docker
+~~~~~~
+
+- ``docker pull ethpm/ethpm:latest``
+- ``docker run ethpm/ethpm:latest``
+
+Homebrew (recommended)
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -1,9 +1,42 @@
...
 
-``pip install ethpm-cli``
+
+Docker
+~~~~~~
+
+- ``docker pull ethpm/ethpm:latest``
+- ``docker run ethpm/ethpm:latest``
+
+Homebrew
```
pipermerriam: Any reason why this one isn't the recommended one?

pipermerriam (approved):

docs/installation.rst
njgheorghita: None at all - just went with pypi to encourage using more python things. Do you think Homebrew should be recommended?

pipermerriam: I think so.  It makes it less of a python tool and more of a general purpose tool which I think is a good thing, and it installs into the global namespace rather than whatever venv they have active (or at least it should...?).

Thoughts?  Ideally, our brew install would install itself into it's own venv (does it do that?)

pipermerriam: similarly, it would be nice to get an `apt-get` install path available soon too for all the linux folks.

njgheorghita: > (does it do that?)
yup!

Added an `apt-get` install path to issue #31

PR #66, opened 2019-09-18
Add amend flag to manifest wizard

njgheorghita:
## What was wrong?
Adding a feature to the manifest wizard to simply amend existing manifests is something that's pretty  useful, i've found. Amending a manifest is currently limited to the meta fields and adding a single deployment. Multiple deployments gets tricky b/c the blockchain URI - but it's on my todo list to support that soon.

Possible extensions could include importing manifests from uris to amend.

#### Cute Animal Picture
image


ethpm_cli/config.py
```diff
@@ -81,6 +81,15 @@ def __init__(self, args: Namespace) -> None:
         else:
             self.project_dir = None
 
+        if "manifest_path" in args and args.manifest_path:
+            if not args.manifest_path.is_file():
+                raise ConfigurationError(
```
pipermerriam, line 86: I think you can do this at the parser level using this pattern

https://github.com/ethereum/trinity/blob/fc194a64215ccca1436f2ad874fcd244ce03db1f/trinity/cli_parser.py#L249-L279

ethpm_cli/manifest.py
```diff
@@ -76,6 +75,168 @@ def generate_custom_manifest(project_dir: Path) -> None:
...
+    builder_fns = (
+        amend_description(manifest),
+        amend_license(manifest),
+        amend_authors(manifest),
+        amend_keywords(manifest),
+        amend_links(manifest),
+        *amend_deployments(manifest),
+        gen_validate_manifest(),
+    )
+    final_fns = (fn for fn in builder_fns if fn is not None)
```
pipermerriam, line 97: Maybe instead of having the functions return `None` they could just return a `noop` function that did nothing.

ethpm_cli/manifest.py
```diff
@@ -76,6 +75,168 @@ def generate_custom_manifest(project_dir: Path) -> None:
...
+    amended_manifest = b.build(manifest, *final_fns)
+    new_filename = input("Please enter a new filename for your manifest. ")
+    new_path = manifest_path.parent / f"{new_filename}.json"
+    while new_path.exists():
```
pipermerriam: I think you can structure this as:

```python
while True:
    file_path = input(...)
    if file_path.exists():
        # sorry a file already exists ....
        continue
    else:
        break
```

ethpm_cli/manifest.py
```diff
@@ -76,6 +75,168 @@ def generate_custom_manifest(project_dir: Path) -> None:
...
+        )
+        return tuple()
+
+    if flag:
+        deployment_data = amend_single_deployment(manifest)
+        return (b.deployment(**dep) for dep in deployment_data)
+    return tuple()
+
+
+@to_tuple
```
pipermerriam, line 221: This looks wrong.  The `yield` below returns a dict, and then I'm pretty sure the `to_tuple` is going to convert that to be only the dict keys in a tuple.

Suggests that there may be missing test coverage.

pipermerriam (approved):

ethpm_cli/config.py
njgheorghita: Is there an advantage to raising it at the parser level? Just curious since it seems like it'll require some extra configuration

ethpm_cli/manifest.py
njgheorghita: To eliminate having to do this entire comprehension and feed the `builder_fns` straight into the build on line 98? I tested with a simple noop and it'll break the manifest builder, so filtering out the `noop`s / `None` will still be necessary - or I can add a special `noop` to `ethpm`'s `builder`

ethpm_cli/manifest.py
njgheorghita: It's working as expected - returns `({},)` - though the test coverage could be better, but there's other improvements to the whole deployments strategy i'm planning to add in a future pr

pipermerriam: I misread what was going on, carry on.

ethpm_cli/config.py
pipermerriam: I believe the advantage is that this is a parser level error and there's a framework already in place via argparse for validating the user input and raising errors.  It's not necessarily better, but it can be good to have isolation, colocating all of a certain phase of validation in the same conceptual part of the application.

ethpm_cli/manifest.py
pipermerriam: Not *just* to eliminate the comprehension, but more to remove the need for all those functions to have `Optional` return types.  Homogeneous typing such that we don't have to have runtime filters like this can be a nice improvement and helps reduce complexity (this comprehension and the `Optional` return types are a type of complexity).

> or I can add a special `noop`

Yes, wasn't necessarily meaning `noop = lambda x: None` but whatever a `noop` would mean in the context of the builder pipeline, a function that does nothing in the pipeline.

PR #68, opened 2019-09-24
Implement examine manifest flag

### LINKED ISSUE #67: Add examine flag to CLI
`ethpm examine --manifest-path [PATH_TO_MANIFEST]`

For a quick & dirty pretty print of the manifest - with truncated fields for easy viewing - could be a helpful feature. Should support both local manifest paths and manifest uris.

njgheorghita:
## What was wrong?
Fixes #67 

## How was it fixed?
Implemented `ethpm examine --manifest-path [PATH_TO_MANIFEST]`

#### Cute Animal Picture
image

### DIFF
--- a/ethpm_cli/manifest.py
+++ b/ethpm_cli/manifest.py
@@ -470,7 +470,7 @@ def parse_bool_flag(question: str) -> bool:
             continue
 
 
-def examine_manifest(manifest_path: Path) -> None:
+def cat_manifest(manifest_path: Path) -> None:
     raw_manifest = json.loads(manifest_path.read_text())
     validate_manifest_against_schema(raw_manifest)
     manifest = ManifestDisplay(raw_manifest)
@@ -502,7 +502,7 @@ class ManifestDisplay:
     @to_list
     def sources(self) -> Iterable[str]:
         if "sources" not in self.manifest:
-            yield "No sources found.\n"
+            yield "None.\n"
         else:
             for src, data in self.manifest["sources"].items():
                 if len(data) < 50:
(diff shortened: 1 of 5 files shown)


ethpm_cli/manifest.py
```diff
@@ -468,3 +468,76 @@ def parse_bool_flag(question: str) -> bool:
...
+        return self.manifest["version"]
+
+    @property
+    def manifest_version(self) -> str:
+        return self.manifest["manifest_version"]
+
+    @to_list
+    def sources(self) -> Iterable[str]:
+        if "sources" not in self.manifest:
+            yield "No sources found.\n"
```
pipermerriam: Maybe just `N/A` or `None`.  The current message kind of sounds like they might be there and just be hidden somewhere..  Same with the other ones that use the same verbage.

ethpm_cli/parser.py
```diff
@@ -512,3 +513,27 @@ def list_action(args: argparse.Namespace) -> None:
...
+    config = Config(args)
+    examine_manifest(config.manifest_path)
+
+
+examine_parser = ethpm_parser.add_parser(
+    "examine",
+    help="Examine the contents of the manifest located at the provided manifest path.",
+)
+examine_parser.add_argument(
+    "--manifest-path",
```
pipermerriam: Might be good to use a positional argument here instead of an option so that this command can be called as `ethpm examine ./path/to/manifest.json`

pipermerriam (approved):
Maybe use unix terminology and change the CLI to be `ethpm cat --manifest-path ...`?  I've seen this convention elsewhere (not exactly sure where...)

PR #69, opened 2019-09-26
Move command files to commands/

njgheorghita:
## What was wrong?
- moved all command specific files to a command directory
- minor refactor in writing new manifest to disk

#### Cute Animal Picture
image


pipermerriam (approved):
rubber stampy, looks like this is mostly code-reorg

PR #70, opened 2019-10-30
Implement ethpm update

njgheorghita:
## What was wrong?
A cli command to update the version of an installed package.

1. Package must be available on the active registry
2. Package install data stored in local lockfile must match what's found on active registry
3. User must input the target version to update to (since there's no requirement for semver, it's hard to infer the newest version)

#### Cute Animal Picture
image

### DIFF
--- a/ethpm_cli/commands/install.py
+++ b/ethpm_cli/commands/install.py
@@ -26,7 +26,7 @@ from ethpm_cli.constants import (
     SRC_DIR_NAME,
 )
 from ethpm_cli.exceptions import InstallError
-from ethpm_cli.validation import validate_parent_directory
+from ethpm_cli.validation import validate_parent_directory, validate_same_registry
 
 logger = logging.getLogger("ethpm_cli.install")
 
@@ -129,7 +129,7 @@ def get_package_aliases(package_name: str, config: Config) -> Iterable[Tuple[str
                 yield alias
 
 
-def resolve_pkg_name(pkg_name: str, config: Config) -> InstalledPackage:
+def resolve_pkg_name_or_alias(pkg_name: str, config: Config) -> InstalledPackage:
     lockfile_path = config.ethpm_dir / "ethpm.lock"
     lockfile = json.loads(lockfile_path.read_text())
     return InstalledPackage(**lockfile[pkg_name])
@@ -137,21 +137,13 @@ def resolve_pkg_name(pkg_name: str, config: Config) -> InstalledPackage:
 
 def update_package(args: Namespace, config: Config) -> None:
     if is_package_installed(args.package, config):
-        installed_pkg = resolve_pkg_name(args.package, config)
+        installed_pkg = resolve_pkg_name_or_alias(args.package, config)
         active_registry = get_active_registry(config.xdg_ethpmcli_root / REGISTRY_STORE)
-        active_registry_uri = parse_registry_uri(active_registry.uri)
         if is_valid_registry_uri(installed_pkg.install_uri):
-            install_uri = parse_registry_uri(installed_pkg.install_uri)
-            if (
-                install_uri.address != active_registry_uri.address
-                or install_uri.chain_id != active_registry_uri.chain_id  # noqa: W503
-            ):
-                raise InstallError(
-                    f"Active registry: {active_registry.uri} does not match the registry URI used "
-                    f"to install {installed_pkg.resolved_package_name}:{installed_pkg.install_uri}."
-                )
+            validate_same_registry(installed_pkg.install_uri, active_registry.uri)
 
         connected_chain_id = config.w3.eth.chainId
+        active_registry_uri = parse_registry_uri(active_registry.uri)
         if not to_int(text=active_registry_uri.chain_id) == connected_chain_id:
             raise InstallError(
                 f"Registry URI chain: {active_registry_uri.chain_id} doesn't match "
--- a/ethpm_cli/commands/install.py
+++ b/ethpm_cli/commands/install.py
@@ -31,24 +31,24 @@ from ethpm_cli.validation import validate_parent_directory, validate_same_regist
 logger = logging.getLogger("ethpm_cli.install")
 
 
-def install_package(pkg: Package, config: Config) -> None:
-    if is_package_installed(pkg.alias, config):
+def install_package(package: Package, config: Config) -> None:
+    if is_package_installed(package.alias, config):
         raise InstallError(
-            f"Installation conflict: Package: '{pkg.manifest['package_name']}' "
-            f"aliased to '{pkg.alias}' already installed on the filesystem at "
-            f"{config.ethpm_dir / pkg.alias}. Try installing this package with "
+            f"Installation conflict: Package: '{package.manifest['package_name']}' "
+            f"aliased to '{package.alias}' already installed on the filesystem at "
+            f"{config.ethpm_dir / package.alias}. Try installing this package with "
             "a different alias."
         )
 
     # Create temporary package directory
-    tmp_pkg_dir = Path(tempfile.mkdtemp())
-    write_pkg_installation_files(pkg, tmp_pkg_dir, config.ipfs_backend)
+    tmp_package_dir = Path(tempfile.mkdtemp())
+    write_package_installation_files(package, tmp_package_dir, config.ipfs_backend)
 
     # Copy temp package directory to ethpm dir namespace
-    dest_pkg_dir = config.ethpm_dir / pkg.alias
-    validate_parent_directory(config.ethpm_dir, dest_pkg_dir)
-    shutil.copytree(tmp_pkg_dir, dest_pkg_dir)
-    install_to_ethpm_lock(pkg, (config.ethpm_dir / LOCKFILE_NAME))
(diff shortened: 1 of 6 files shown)


ethpm_cli/commands/install.py
```diff
@@ -119,6 +129,101 @@ def get_package_aliases(package_name: str, config: Config) -> Iterable[Tuple[str
                 yield alias
 
 
+def resolve_pkg_name(pkg_name: str, config: Config) -> InstalledPackage:
```
pipermerriam: Maybe `resolve_installed_package_by_name`?  It took me reading the body of the function to understand what this was doing, i.e. looking at the currently installed packages and grabbing the one specified by the given name.

ethpm_cli/commands/install.py
```diff
@@ -119,6 +129,93 @@ def get_package_aliases(package_name: str, config: Config) -> Iterable[Tuple[str
                 yield alias
 
 
+def resolve_pkg_name_or_alias(pkg_name: str, config: Config) -> InstalledPackage:
```
pipermerriam: I'd suggest two things.

1. Don't abbreviate and use `_package_` and stick with that through all of the naming you do as often as possible.  It will alleviate confusion and mental overhead if you just stick with the long form and thus, never have to remember which ones use the abbreviated version and which don't.
2. Maybe name it `resolve_installed_package` which I think better communicates what this is doing.  On my first read I had to dig into the function body to figure it out as it wasn't clear what resolve meant.  (this also implies renaming the argument `pkg_name -> package_name_or_alias` or maybe `package_identifier` if you want to use that naming convention to represent *either*.

ethpm_cli/commands/install.py
```diff
@@ -119,6 +129,93 @@ def get_package_aliases(package_name: str, config: Config) -> Iterable[Tuple[str
...
+    lockfile_path = config.ethpm_dir / "ethpm.lock"
+    lockfile = json.loads(lockfile_path.read_text())
+    return InstalledPackage(**lockfile[pkg_name])
```
pipermerriam: You *might* check what mypy says about `reveal_type(lockfile)`.  This is more of a nitpick to try and get more confidence in the types for instances of `InstalledPackage` but you *might* consider something like this as I've found that `mypy` doesn't necesarily get along well with `**kwargs` style code.

```python
Lockfile = TypedDict(...) define the types for a lockfile dictionary somewhere.

class InstalledPackage:
    @classmethod
    def from_lockfile(cls, lockfile: Lockfile) -> 'InstalledPackage':
        return cls(
            arg1=lockfile['arg1'],  # these will be strongly typed due to the `TypedDict`
            ...
        )
```

This may not actually be a very big improvement and *maybe* mypy is doing better with this stuff so feel free to disregard this entirely or do some checking with `reveal_type(...)` to figure out if there's even a problem with the types.

ethpm_cli/commands/install.py
```diff
@@ -119,6 +129,93 @@ def get_package_aliases(package_name: str, config: Config) -> Iterable[Tuple[str
...
+
+
+def update_package(args: Namespace, config: Config) -> None:
+    if is_package_installed(args.package, config):
```
pipermerriam: You might invert this so that you do an early exit in the event of what is currently the `else` clause, and then you can dedent this entire large block of code under the `if`.  I'd call this a minor *code smell*, having such a huge block of code inside of an `if` statement like this.

ethpm_cli/commands/install.py
```diff
@@ -119,6 +129,93 @@ def get_package_aliases(package_name: str, config: Config) -> Iterable[Tuple[str
...
+                f"Install URI found on active registry for {installed_pkg.resolved_package_name}@"
+                f"{installed_pkg.resolved_version}: {on_chain_install_uri} does not match the "
+                f"install URI found in local lockfile: {installed_pkg.resolved_uri}."
+            )
+
+        cli_logger.info(
+            f"{len(all_versions)} versions of {installed_pkg.resolved_package_name} "
+            f"found on the active registry: {all_versions}."
+        )
+        while True:
```
pipermerriam: I'd put an upper bound on the number of iterations here and you can error out a final `else` clause.  Generally good to avoid infinite loops.

ethpm_cli/commands/install.py
```diff
@@ -119,6 +129,93 @@ def get_package_aliases(package_name: str, config: Config) -> Iterable[Tuple[str
...
+        # Create an updated args/Package for new install
+        updated_args = copy.deepcopy(args)
+        if installed_pkg.resolved_package_name != args.package:
+            updated_args.alias = args.package
+        updated_args.uri = pluck_release_data(all_release_data, target_version)
+        updated_args.package_version = target_version
+        updated_package = Package(updated_args, config.ipfs_backend)
+
+        uninstall_package(args.package, config)
+        install_package(updated_package, config)
```
pipermerriam: I think you need to make this operation and the previous one atomic.  If the program errors out between lines 203/204 then you'll end up in an inconsistent state with the previous package uninstalled but the new one missing.

tests/core/test_validation.py
```diff
@@ -85,3 +85,48 @@ def test_validate_install_cli_args_rejects_invalid_relative_paths(
...
+@pytest.mark.parametrize(
+    "left,right",
+    (
+        (
+            "erc1319://0x6b5DA3cA4286Baa7fBaf64EEEE1834C7d430B729:1",
+            "erc1319://0x6b5DA3cA4286Baa7fBaf64EEEE1834C7d430B729:1",
+        ),
+        (
+            "erc1319://0x6b5DA3cA4286Baa7fBaf64EEEE1834C7d430B729:1/dai?version=1.0.0",
+            "erc1319://0x6b5DA3cA4286Baa7fBaf64EEEE1834C7d430B729:1",
```
pipermerriam, line 99: do we support ENS names in these?

pipermerriam (dismissed):

tests/core/test_validation.py
njgheorghita: in theory yes, in reality, not until I make an update to ethpm, but that pr is coming soon

njgheorghita:
@pipermerriam ping for another :eyes:

ethpm_cli/commands/install.py
```diff
@@ -111,25 +121,132 @@ def get_package_aliases(package_name: str, config: Config) -> Iterable[Tuple[str
...
+        check_for_aliased_package(package_name, config)
         return
 
+    tmp_package_dir = Path(tempfile.mkdtemp()) / ETHPM_PACKAGES_DIR
+    shutil.copytree(config.ethpm_dir, tmp_package_dir)
+    shutil.rmtree(tmp_package_dir / package_name)
+    uninstall_from_ethpm_lock(package_name, (tmp_package_dir / LOCKFILE_NAME))
+
+    shutil.rmtree(config.ethpm_dir)
+    tmp_package_dir.replace(config.ethpm_dir)
```
njgheorghita: Is it correct that this isn't atomic and needs a context manager? If something happens b/w 245/46 then we just end up with a missing `ethpm_dir`?

pipermerriam: Yes, but I think you can fix this simply by just removing line 245 since `tmp_package_dir.replace(...)` will overwrite the target directory (thus no need to remove it beforehand).

https://docs.python.org/3/library/pathlib.html#pathlib.Path.replace

ethpm_cli/commands/install.py
```diff
@@ -111,25 +121,132 @@ def get_package_aliases(package_name: str, config: Config) -> Iterable[Tuple[str
...
+    # Create an updated args/Package for new install
+    updated_args = copy.deepcopy(args)
+    if installed_package.resolved_package_name != args.package:
+        updated_args.alias = args.package
+    updated_args.uri = pluck_release_data(all_release_data, target_version)
+    updated_args.package_version = target_version
+    updated_package = Package(updated_args, config.ipfs_backend)
+
+    # atomic replace
+    with tempfile.TemporaryDirectory() as tmpdir:
```
pipermerriam, line 214: You might be able to wrap this logic up as something like:

```python
with config.as_temporary_dir() as tmp_config:
    # here `tmp_config.ethpm_dir` is the temporary directory
# upon context exit it does the `tmp_config.ethpm_dir.replace(config.ethpm_dir)` to make the updates atomic.

pipermerriam (approved):

ethpm_cli/commands/install.py
njgheorghita: Yup, I read that and tried, but for some reason when just using `.replace(...)` without the `shutil.rmtree(...)` line - even though the docs say `If target points to an existing file or directory, it will be unconditionally replaced.` - I get an error 

> OSError: [Errno 66] Directory not empty: '/var/folders/w8/9nyrlhw90zgb_cyhtnz5gzhc0000gn/T/tmpo_50cgis/_ethpm_packages' -> '/Users/nickgheorghita/ethereum/ethpm-cli/_ethpm_packages'

pipermerriam: Ah, I think it's the same way that you can't blindly remove a non-empty directory in Unix.  Maybe worth leaving a comment here to make it clear why they are both necessary.

PR #71, opened 2019-10-31
Add option to alias deployments in manifest wizard

njgheorghita:
## What was wrong?
Option to alias deployments was needed in the manifest wizard.

#### Cute Animal Picture
image

### DIFF
--- a/ethpm_cli/commands/manifest.py
+++ b/ethpm_cli/commands/manifest.py
@@ -209,8 +209,7 @@ def amend_single_deployment(manifest: Dict[str, Any]) -> Iterable[Dict[str, Any]
     address = get_deployment_address()
     available_contract_types = manifest["contract_types"].keys()
     contract_type = get_deployment_contract_type(available_contract_types)
-    # todo: support custom contract instance definitions
-    contract_instance = contract_type
+    contract_instance = get_deployment_alias(contract_type)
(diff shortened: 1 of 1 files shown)


pipermerriam (approved):
I know they are likely cumbersome but I think it *might* be worth a test.

If you're finding the tests hard to write that probably suggests thinking about how you can make tests easier to write.  In our other projects, we've tended to have good results in declarative tooling that captures the high level *actions* you want to perform using concise syntax/apis.  For this it *might* mean some sort of simple control flow like:

```python
do_cli(
    command(install).option('--some-option'),  # some sort of simple way to construct the cli command
    assert_text(...),  # simple assertions about the CLI outputting text
    enter_text('y'),  # respond to a prompt
    assert_lockfile(
        has_package_name('owned', version='1.0.0'),
        has_aliased_package('owned', 'owned-alias'),
    ),
    assert_has_directory(...),
)
```

These would all be pretty simple utility functions to write but it might improve your ability to test complex stuff without having to rebake the same tedious blocks of code over and over.

njgheorghita:
@pipermerriam Yup, testing has gotten pretty tricky - `pexpect`'s been pretty great so far, but something along the API you suggested looks really useful, i'll experiment with implementing something like that in a later pr

PR #75, opened 2020-01-13
Write automatic solidity compilation

njgheorghita:
## What was wrong?
`ethpm create manifest-wizard` will now automatically compile your contracts if it can't find the `solc_output` in the projects folder and it detects an available compiler.

#### Cute Animal Picture
image

### DIFF
--- a/ethpm_cli/_utils/solc.py
+++ b/ethpm_cli/_utils/solc.py
@@ -1,5 +1,8 @@
 import json
 from pathlib import Path
+import shutil
+import subprocess
+import tempfile
 from typing import Any, Callable, Dict, Iterable, Tuple
 
 from eth_typing import Manifest
@@ -10,6 +13,7 @@ from ethpm.tools import builder as b
 from ethpm_cli._utils.ipfs import get_ipfs_backend
 from ethpm_cli._utils.logger import cli_logger
 from ethpm_cli.constants import SOLC_INPUT, SOLC_OUTPUT
+from ethpm_cli.exceptions import CompilationError
 
 BASE_SOLC_INPUT = {
     "language": "Solidity",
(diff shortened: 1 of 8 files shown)


ethpm_cli/_utils/solc.py
```diff
@@ -43,12 +47,53 @@ def generate_solc_input(contracts_dir: Path) -> None:
...
         SOLC_INPUT,
     )
-    cli_logger.info(
-        f"Use `solc --standard-json --allow-paths /[ABS_PATH_TO]/{contracts_dir} "
-        f"< /[REL_PATH_TO]/{SOLC_INPUT} > /[REL_PATH_TO]/{contracts_dir.parent / SOLC_OUTPUT}` "
-        "to generate the Solidity compiler output. "
-        "Requires that you have the correct Solidity compiler version installed."
+
+
+def compile_contracts(project_dir: Path) -> None:
```
pipermerriam: I *might* recommend splitting this into two functions.

```python
def find_contracts(root_dir: Path) -> Tuple[Path, ...]:
    ...

def find_solidity_compiler() -> Path:
    ...

def compile_contracts(solc_path: Path: root_dir: Path, source_paths: Collection[Path]) -> None:
    ...
```

This decouples finding the sources from the compilation which are two somewhat distinct phases of the process.  This code is probably fine as is, but I wanted to point this out as a general architectural pattern to look out for.

One small benefit is that you can then decouple the `INFO` level logging from the low level functions.

```python
def cli_compile_comtracts(...):
    paths_to_compile = find_contracts(...)
    logger.info("Compiling contracts: ...")
    if 'ETHPM_SOLC_PATH' in os.environ:
        ...
    else:
        try:
            solc_path = find_solc_path()
            logger.debug("Using solc compiler executable: %s', solc_compiler)
        FileNotFoundError:
            logger.info("No solidity compiler found...")
    compile_contracts(solc_path, ...)
    logger.info("Finished compiling, ...")
```

ethpm_cli/_utils/solc.py
```diff
@@ -43,12 +47,53 @@ def generate_solc_input(contracts_dir: Path) -> None:
...
+    contracts_dir = project_dir / "contracts"
+    contracts = [contract.name for contract in contracts_dir.glob("**/*.sol")]
+    if len(contracts) == 0:
+        raise CompilationError(f"No Solidity contracts found in {contracts_dir}.")
+    contracts_display = "\n- ".join(contracts)
+    cli_logger.info("Compiling contracts found in %s", contracts_dir)
+    cli_logger.info("%d contracts found: \n- %s", len(contracts), contracts_display)
+
+    tmp_project_dir = Path(tempfile.TemporaryDirectory().name) / "project"
```
pipermerriam: You probably want to be using the context manager interface for the temporary directory so that it automatically gets cleaned up.

pipermerriam (approved):

PR #76, opened 2020-01-27
Design

njgheorghita:
## What was wrong?

Changed CLI Command
`ethpm create manifest-wizard ....` -> `ethpm create wizard ...`

Added a logo ... but I'm not sure if it's overkill to have this appear for every cli command usage...

Added some color to various ethpm command outputs

Fixed test config so cli tests now run in circle

#### Cute Animal Picture
image

### DIFF
--- a/ethpm_cli/commands/etherscan.py
+++ b/ethpm_cli/commands/etherscan.py
@@ -46,7 +46,6 @@ def build_etherscan_manifest(
     w3 = setup_w3(to_int(text=chain_id))
     block_uri = create_latest_block_uri(w3)
     runtime_bytecode = to_hex(w3.eth.getCode(address))
-    print(body)
 
     yield "package_name", package_name
     yield "version", version
@@ -83,8 +82,10 @@ def make_etherscan_request(contract_addr: str, network: str) -> Dict[str, Any]:
 def parse_etherscan_response(
     response: Dict[str, Any], contract_addr: str
 ) -> Dict[str, Any]:
-    unverified_msg = 'Contract source code not verified'
-    unverified_contracts = [res for res in response['result'] if res['ABI'] == unverified_msg]
+    unverified_msg = "Contract source code not verified"
+    unverified_contracts = [
+        res for res in response["result"] if res["ABI"] == unverified_msg
+    ]
     if response["message"] == "NOTOK" or unverified_contracts:
         raise ContractNotVerified(
             f"Contract at {contract_addr} unavailable or has not been verified on Etherscan."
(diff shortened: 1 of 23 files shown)


ethpm_cli/commands/etherscan.py
```diff
@@ -82,7 +82,11 @@ def make_etherscan_request(contract_addr: str, network: str) -> Dict[str, Any]:
 def parse_etherscan_response(
     response: Dict[str, Any], contract_addr: str
 ) -> Dict[str, Any]:
-    if response["message"] == "NOTOK":
+    unverified_msg = "Contract source code not verified"
```
pipermerriam: This can be a module level constant.

ethpm_cli/commands/install.py
```diff
@@ -73,7 +76,9 @@ def format_for_display(self) -> str:
             alias = f" (alias: {self.path.name})"
         else:
             alias = ""
-        main_info = f"{prefix}{self.package_name}{alias}=={self.package_version}"
+        blue_package_name = f"{BLUE_STRING}{self.package_name}{COLOR_RESET}"
```
pipermerriam: I'd recommend utility functions for this.

```python
def blue(text):
    return f"{BLUE_STRING}{text}{COLOR_RESET}"
```
blue_package_name = blue(self.package_name)

Or you can use the `termcolor` package like we do in trinity: https://github.com/ethereum/trinity/blob/master/trinity/_utils/shellart.py

ethpm_cli/main.py
```diff
@@ -1,10 +1,28 @@
...
 from ethpm_cli.parser import parser
 
+ETHPM_LOGO = """
+       _   _     ____  __  __
+   ___| |_| |__ |  _ \|  \/  |
+  / _ \ __| '_ \| |_) | |\/| |
+ |  __/ |_| | | |  __/| |  | |
+  \___|\__|_| |_|_|   |_|  |_|
+
+"""  # noqa: W605
```
pipermerriam: I'm guessing that the logo is too much for now.  I'd suggest not doing this unless it's common in other package managers (which I don't recall seeing very often).

Maybe this could only be displayed in the `version` or `help` commands (and myabe `wizard`)?

tests/cli/test_create.py
```diff
@@ -2,20 +2,18 @@
...
+from ethpm_cli.constants import SOLC_INPUT
 
 
 def test_custom_manifest_builder(tmp_project_dir, test_assets_dir):
     child = pexpect.spawn(
-        f"ethpm create manifest-wizard --project-dir {tmp_project_dir}", timeout=5
+        f"ethpm create wizard --project-dir {tmp_project_dir}", timeout=5
     )
-    child.expect(f"ethPM CLI v{ETHPM_CLI_VERSION}\r\n")
+    child.expect(f"A command line tool for the Ethereum Package Manager.")
```
pipermerriam: Maybe make this a constant you can import so the text isn't reproduced across all these files.

pipermerriam (approved):

PR #77, opened 2020-01-28
Implement ethpm activate

njgheorghita:
## What was wrong?
`ethpm activate [installed_package_name or uri]` command to make it super duper simple to interact with any package in your terminal.

Automatically opens up an ipython console with all contract factories and deployments found in the package loaded and ready to use.

#### Cute Animal Picture
image

### DIFF
--- a/ethpm_cli/commands/activate.py
+++ b/ethpm_cli/commands/activate.py
@@ -3,7 +3,7 @@ import json
 from urllib import parse
 
 from IPython import embed
-from eth_utils import to_dict
+from eth_utils import to_dict, to_tuple
 from ethpm import Package as ethpmPackage
 from ethpm._utils.chains import parse_BIP122_uri
 from ethpm.exceptions import InsufficientAssetsError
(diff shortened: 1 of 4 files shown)


ethpm_cli/commands/activate.py
```diff
@@ -0,0 +1,142 @@
...
+        if not is_package_installed(args.package_or_uri, config):
+            raise InstallError(
+                f"Package: {args.package_or_uri} not installed in ethPM dir: {config.ethpm_dir}."
+            )
+        manifest = json.loads(
+            (config.ethpm_dir / args.package_or_uri / "manifest.json").read_text()
+        )
+
+    pkg = ethpmPackage(manifest, config.w3)
+    cli_logger.info("\U000026A1" * 3)
```
pipermerriam: These unicode characters deserve a comment to say what they are.

ethpm_cli/commands/activate.py
```diff
@@ -0,0 +1,142 @@
...
+    else:
+        cli_logger.info(f"No deployments found.\n")
+        available_deployments = {}
+
+    # instantiate contract factory variables
+    if len(available_factories) > 0:
+        cli_logger.info(f"Generated {len(available_factories)} contract factories: ")
+        for key, val in available_factories.items():
+            cli_logger.info(f"- {key}_factory")
+            exec(f"{key}_factory" + "=val")
```
pipermerriam: This is a (potential) remote code execution vulnerability.  In theory if a package contained a `key` that was python code then this would execute it on the user's machine.  I think our constraints on the allowed values prevent this but it's still not something we should have in our code.

Rather than *injecting* these into the local namespace, you should instead do something like what we do here in the `trinity attach` command.

https://github.com/ethereum/trinity/blob/master/trinity/components/builtin/attach/console.py

See the `user_ns` object that is built to contain all of the values the user might want.  You can also build up a text header instead of using logging to display what variables are available for use.

setup.py
```diff
@@ -52,6 +51,7 @@
     include_package_data=True,
     install_requires=[
         "eth-hash[pysha3]>=0.2.0,<1",
+        "ipython",
```
pipermerriam: This should be pinned to a version range.

pipermerriam (approved):

Issue #19, opened 2019-05-16
Refactor `get_ethpm_birth_block`


njgheorghita:
`get_ethpm_birth_block` should be refactored to use a non-recursive solution to find the target block. re: https://github.com/ethpm/ethpm-cli/pull/16#discussion_r284329903

PR #78, opened 2020-02-05
Refactor to non-recursive solution for target block

corydickson:
## What was wrong?

Issue #19 

## How was it fixed?

The approach laid out here attempts to reach the target block by measuring the time passed between initial block and the last block given. It is considerably slower than the recursive solution, though addresses the implementation concern. Most likely I'm missing something by not taking advantage of the fact we can look at the block in-between this initial range and search on a smaller set of inputs?

#### Cute Animal Picture

### DIFF
--- a/ethpm_cli/commands/scraper.py
+++ b/ethpm_cli/commands/scraper.py
@@ -85,7 +85,11 @@ def get_ethpm_birth_block(
             from_block = mid + 1
         else:
             return mid - 1
-    return -1
+
+    raise BlockNotFoundError(
+        f"Cannot find closest block to timestamp: {target_timestamp} "
+        f"in range given {from_block} - {to_block}."
+    )
 
 
 def block_range_needs_scraping(
--- a/ethpm_cli/commands/scraper.py
+++ b/ethpm_cli/commands/scraper.py
@@ -76,34 +76,16 @@ def get_ethpm_birth_block(
     """
     version_release_date = datetime.fromtimestamp(target_timestamp)
 
-    mid_block = int((from_block + to_block) / 2)
-
-    from_date = datetime.fromtimestamp(w3.eth.getBlock(from_block)["timestamp"])
-    to_date = datetime.fromtimestamp(w3.eth.getBlock(to_block)["timestamp"])
-    mid_date = datetime.fromtimestamp(w3.eth.getBlock(mid_block)["timestamp"])
-
-    delta_start = version_release_date - from_date
-    delta_end = version_release_date - to_date
-    delta_mid = version_release_date - mid_date
-
-    min_delta = min(delta_mid.days, delta_start.days, abs(delta_end.days))
-
-    if min_delta == abs(delta_end.days):
-        while to_date < version_release_date:
-            to_block -= 1
-            to_date = datetime.fromtimestamp(w3.eth.getBlock(to_block)["timestamp"])
-        return to_block + 1
-
-    elif min_delta == delta_mid.days:
-        while mid_date < version_release_date:
-            mid_block += 1
-            mid_date = datetime.fromtimestamp(w3.eth.getBlock(mid_block)["timestamp"])
-        return mid_block - 1
-    else:
-        while version_release_date < mid_date:
-            mid_block -= 1
-            mid_date = datetime.fromtimestamp(w3.eth.getBlock(mid_block)["timestamp"])
-        return mid_block + 1
+    while from_block < to_block:
+        mid = (from_block + to_block) // 2
+        target = datetime.fromtimestamp(w3.eth.getBlock(mid)["timestamp"])
+        if target > version_release_date:
+            to_block = mid
+        elif target < version_release_date:
+            from_block = mid + 1
+        else:
+            return mid - 1
+    return -1
 
 
 def block_range_needs_scraping(
--- a/ethpm_cli/commands/scraper.py
+++ b/ethpm_cli/commands/scraper.py
@@ -76,22 +76,27 @@ def get_ethpm_birth_block(
     """
     version_release_date = datetime.fromtimestamp(target_timestamp)
     from_date = datetime.fromtimestamp(w3.eth.getBlock(from_block)["timestamp"])
-    delta = version_release_date - from_date
+    to_date = datetime.fromtimestamp(w3.eth.getBlock(to_block)["timestamp"])
 
-    if delta.days <= 0 and from_date < version_release_date:
-        while (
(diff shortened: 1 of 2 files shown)


njgheorghita:
@corydickson Thanks for the pr! The broken tests are b/c of web3's latest update - I've got a fix coming soon that should solve them

njgheorghita:
Fix is in - if you rebase against master that should clear up the failing tests

corydickson:
@njgheorghita Thank you, tho I'm having trouble reproducing the failure in the latest build above to fix


```
============================================================================================ slowest 10 test durations =============================================================================================
53.84s call     tests/core/test_scraper.py::test_scraper_imports_existing_ethpmcli_dir
41.02s call     tests/core/test_scraper.py::test_scraper_writes_to_disk
26.66s call     tests/core/test_scraper.py::test_get_ethpm_birth_block[4000]
17.98s call     tests/cli/test_registry_cli.py::test_ethpm_registry_commands
13.51s call     tests/cli/test_activate.py::test_activate_etherscan_uri_with_single_deployment
12.91s call     tests/cli/test_install_cli.py::test_ethpm_install_from_etherscan
11.90s call     tests/cli/test_activate.py::test_activate_ipfs_uri_with_factories_and_deployments
10.51s call     tests/cli/test_activate.py::test_activate_registry_uri_with_contract_types_no_deployments
7.12s call     tests/core/test_install.py::test_uninstall_packages[owned-wallet]
6.97s call     tests/core/test_install.py::test_install_multiple_packages
==================================================================================== 120 passed, 1 warning in 627.24s (0:10:27) ====================================================================================

File: ethpm_cli/commands/scraper.py [hunk 1]
```diff
@@ -75,23 +75,35 @@ def get_ethpm_birth_block(
...
 
+    from_date = datetime.fromtimestamp(w3.eth.getBlock(from_block)["timestamp"])
+    to_date = datetime.fromtimestamp(w3.eth.getBlock(to_block)["timestamp"])
+    mid_date = datetime.fromtimestamp(w3.eth.getBlock(mid_block)["timestamp"])
+
+    delta_start = version_release_date - from_date
+    delta_end = version_release_date - to_date
+    delta_mid = version_release_date - mid_date
+
+    min_delta = min(delta_mid.days, delta_start.days, abs(delta_end.days))
```
njgheorghita: I'm a little confused by what's going on here. Wouldn't the `min_delta` always be `delta_end.days`?

It also seems that adjusting the hunt for the next block by one is going to take a looong time. I'd try looking at [this](https://www.sanfoundry.com/python-program-implement-binary-search-without-recursion/) and see if you can get some better performance by following that pattern.

corydickson: You're right, thanks for that resource. I see now that you have to update the range you're looking at for each iteration 😓

tests/core/test_scraper.py
```diff
@@ -166,6 +166,8 @@ def test_get_ethpm_birth_block(w3, interval):
     time_travel(w3, interval)
     actual = get_ethpm_birth_block(w3, 0, w3.eth.blockNumber, latest_block.timestamp)
     assert actual == latest_block.number - 1
+    # Not found
+    assert get_ethpm_birth_block(w3, 0, 0, latest_block.timestamp) == -1
```
corydickson: Hope it's okay to add this to the existing test instead of a separate function even though it's no longer looking at only valid intervals

corydickson:
Thank you for the swift review @njgheorghita ! Let me know if you'd like me to squash the previous commits if things look good

ethpm_cli/commands/scraper.py
```diff
@@ -75,23 +75,17 @@ def get_ethpm_birth_block(
...
+    while from_block < to_block:
+        mid = (from_block + to_block) // 2
+        target = datetime.fromtimestamp(w3.eth.getBlock(mid)["timestamp"])
+        if target > version_release_date:
+            to_block = mid
+        elif target < version_release_date:
+            from_block = mid + 1
+        else:
+            return mid - 1
+    return -1
```
pipermerriam: I think this should be an exception case since otherwise it requires checking the return value which is an antipattern since it is easy to forget to do.

tests/core/test_scraper.py
njgheorghita: It'd be better if it went in its own test - testing that it raises an exception (a la piper's comment). In most cases, it's better to separate testing concerns into as many functions as necessary rather than squeezing them into a single test (though I'm definitely guilty of this also)

tests/core/test_scraper.py
```diff
@@ -168,6 +169,15 @@ def test_get_ethpm_birth_block(w3, interval):
     assert actual == latest_block.number - 1
 
 
+def test_get_ethpm_birth_block_invalid_range_raises_exception(w3):
+    latest_block = w3.eth.getBlock("latest")
+    with pytest.raises(BlockNotFoundError):
+        get_ethpm_birth_block(w3, 0, w3.eth.blockNumber, latest_block.timestamp)
+    with pytest.raises(BlockNotFoundError):
+        time_travel(w3, 1)
+        get_ethpm_birth_block(w3, w3.eth.blockNumber, 0, latest_block.timestamp)
```
corydickson, line 182: For when the `to_block` is strictly larger than the `from_block`

njgheorghita (approved):
Thanks! and ughh these cli tests are so flaky - they're working fine locally, so my guess is there's something funky b/w infura requests and circle.

tests/core/test_scraper.py
njgheorghita: cool! notes like this are definitely worth a comment in the code - and honestly it might be better off as a separate test - then in the test name you can explain what's going on

corydickson: Sure, still trying to discern when it's bloat and when it's appropriate for clarity. Any rule of thumbs you use?

corydickson:
@pipermerriam @njgheorghita Thank you both learned a lot !

tests/core/test_scraper.py
njgheorghita: Hmmm, imo with tests bloat is not such a big concern, I tend to default to the more tests the better. It's good to have only one specific cause/effect being tested per test. We like to use `@pytest.mark.parametrize` ([example](https://github.com/ethpm/ethpm-cli/blob/master/tests/core/test_config.py#L63)) a lot which helps improve the testing quality while reducing bloat.  ping @pipermerriam for thoughts

Issue #60, opened 2019-08-20
Extend ethpm release for local manifests


njgheorghita:
As a user, who has generated a local manifest (i.e. with `ethpm create basic-manifest` or `ethpm create manifest-wizard`) I want to be able to easily release the manifest to my registry by simply providing the path to the local manifest

`ethpm release --manifest-path /path/to/manifest.json --keyfile-password [PASSWORD]`

Command will automatically pin the manifest to ipfs (infura / local /ethpm nodes) & release the manifest with its embedded package name & version to the active registry. 

And improve log msgs for better ux.

PR #81, opened 2020-02-06
Support local manifest to be pinned and released

corydickson:
## What was wrong?

Issue #60 

## How was it fixed?
- [x] Added `--manifest_path` cli option
- [x] Automatically pins file if given
- [x] Updated logging 
- [ ] Test coverage

Need a bit of guidance on where to add tests for this new feature, had trouble finding the tests that cover the default use case. 99% sure this works as intended :) 

Also should I add errors/checks if the `manifest_json` file is malformed in some way? It looks like guards are already in place that ensure it's an existing file.

#### Cute Animal Picture

### DIFF
--- a/ethpm_cli/parser.py
+++ b/ethpm_cli/parser.py
@@ -123,18 +123,21 @@ def add_manifest_path_to_parser(
     )
 
 
-def add_package_name_and_package_version_to_parser(
-    parser: argparse.ArgumentParser, name_msg: str, version_msg: str,
-) -> None:
+def add_package_name_to_parser(parser: argparse.ArgumentParser, help_msg: str,) -> None:
     parser.add_argument(
-        "--package-name", dest="package_name", action="store", type=str, help=name_msg
+        "--package-name", dest="package_name", action="store", type=str, help=help_msg
     )
+
+
+def add_package_version_to_parser(
+    parser: argparse.ArgumentParser, help_msg: str,
+) -> None:
     parser.add_argument(
         "--package-version",
         dest="package_version",
         action="store",
         type=str,
-        help=version_msg,
+        help=help_msg,
     )
 
 
@@ -159,19 +162,17 @@ def release_cmd(args: argparse.Namespace) -> None:
         )
 
         release_package(package_name, package_version, manifest_uri, config)
-        cli_logger.info(
-            f"{package_name} v{package_version} @ {manifest_uri} "
-        )
+        cli_logger.info(f"{package_name} v{package_version} @ {manifest_uri} ")
     else:
-        release_package(args.package_name, args.package_version, args.manifest_uri, config)
+        release_package(
+            args.package_name, args.package_version, args.manifest_uri, config
+        )
         cli_logger.info(
             f"{args.package_name} v{args.package_version} @ {args.manifest_uri} "
         )
 
     active_registry = get_active_registry(config.xdg_ethpmcli_root / REGISTRY_STORE)
-    cli_logger.info(
-        f"released to registry @ {active_registry.uri}."
-    )
+    cli_logger.info(f"released to registry @ {active_registry.uri}.")
 
 
 release_parser = ethpm_parser.add_parser(
@@ -188,9 +189,12 @@ release_group = release_parser.add_mutually_exclusive_group()
 add_manifest_path_to_parser(
     release_group, "Local path to target manifest used for release.",
 )
-add_package_name_and_package_version_to_parser(
+add_package_name_to_parser(
     release_parser,
     "Package name of package you want to release. Must match `package_name` in manifest.",
+)
+add_package_version_to_parser(
+    release_parser,
     "Version of package you want to release. Must match the `version` field in manifest.",
 )
 add_keyfile_password_arg_to_parser(release_parser)
@@ -375,9 +379,12 @@ create_basic_parser = create_subparsers.add_parser(
     "The generated manifest will package up all available sources and contract types "
     "available in the solidity compiler output found in given project directory.",
 )
-add_package_name_and_package_version_to_parser(
+add_package_name_to_parser(
     create_basic_parser,
     "Package name for generating manifest with `basic-manifest` command.",
+)
+add_package_version_to_parser(
+    create_basic_parser,
     "Package version for generating manifest with `basic-manifest` command.",
 )
 add_project_dir_arg_to_parser(create_basic_parser)
@@ -486,9 +493,12 @@ install_parser.add_argument(
     action="store_true",
     help="Flag to use locally running IPFS node rather than defualting to Infura.",
 )
-add_package_name_and_package_version_to_parser(
+add_package_name_to_parser(
     install_parser,
     "Package name to use when installing a package from etherscan URIs.",
+)
+add_package_version_to_parser(
+    install_parser,
     "Package version to use when installing a package from etherscan URIs.",
 )
 add_alias_arg_to_parser(install_parser)
--- a/ethpm_cli/parser.py
+++ b/ethpm_cli/parser.py
@@ -150,11 +150,6 @@ ethpm_parser = parser.add_subparsers(help="CLI commands", dest="command")
 def release_cmd(args: argparse.Namespace) -> None:
     config = Config(args)
 
-    manifest_uri = args.manifest_uri
-
-    package_name = args.package_name
-    package_version = args.package_version
-
     if args.manifest_path:
         (package_name, package_version, manifest_uri) = pin_local_manifest(
             args.manifest_path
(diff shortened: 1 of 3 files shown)


corydickson:
After investigating a bit further, I believe that the issue with adding coverage is that we'll need to set up a local chain to test the `release` cmd properly.

ethpm_cli/parser.py
```diff
@@ -117,13 +120,32 @@ def add_alias_arg_to_parser(parser: argparse.ArgumentParser) -> None:
...
 
-# todo: extend to pin & release a local manifest
 def release_cmd(args: argparse.Namespace) -> None:
     config = Config(args)
-    release_package(args.package_name, args.package_version, args.manifest_uri, config)
+
+    manifest_uri = args.manifest_uri
+
+    package_name = args.package_name
+    package_version = args.package_version
```
njgheorghita: If we're releasing local manifest, then I think we should make the args (`manifest_uri`, `package_name`, `package_version`) and (`manifest_path`) mutually exclusive groups (like we do with `ethpm get` on line 589). We just want to avoid any confusion since the package info found in the manifest & the newly generated hash will override whatever the user passes in.

ethpm_cli/parser.py
```diff
@@ -117,13 +120,32 @@ def add_alias_arg_to_parser(parser: argparse.ArgumentParser) -> None:
...
+
+    if args.manifest_path:
```
njgheorghita, line 156: This logic is great - I'm just not sure this is the best place for it to live. Maybe turn this into something like `pin_local_manifest(manifest_path)  -> package_name, package_version, manifest_uri`. And move it to `ethpm._utils.ipfs` should be a fine home.

There is manifest validation happening in web3. But, it's a good idea to throw in an extra validation check here with `from ethpm.validation.manifest import validate_manifest_against_schema`

ethpm_cli/parser.py
```diff
@@ -152,6 +174,13 @@ def release_cmd(args: argparse.Namespace) -> None:
     type=str,
     help="Content addressed URI at which the manifest for released package is located.",
 )
+release_parser.add_argument(
```
njgheorghita: Once we start having the same flag for multiple commands - I like to move them to helper fns - under `Shared args` at the top of this file. With this addition, there are 3 commands with a `manifest_path` flag, so it'd be nice to move it to a helper function - and you might also have to add an argument for the help message - since it's different between the commands. 

```
def add_manifest_path_arg_to_parser(parser, help_msg):
```

```
add_manifest_path_arg_to_parser(release_parser, "Local path to target manifest used for release.")
```

njgheorghita (changes requested):
Thanks for the pr! Just a couple adjustments I'd like to happen. And you're right about testing. I've always just been testing `release` by locally running commands - which is obviously a horrible way to test 😬 . I'm not sure it's the highest priority, but if you're up for a challenge - adding tests for `release` would be wonderful! The [`pm`](https://github.com/ethereum/web3.py/blob/master/tests/core/pm-module/conftest.py) module in `web3` would be a good place to get an idea on how to get the chain up & running.

ethpm_cli/parser.py
corydickson: Perfect, can this PR also do the same for the other reused args I might come across?

corydickson:
Over the next few days I'm going to take on setting up that testing environment, thank you for the guidance!

ethpm_cli/parser.py
```diff
@@ -108,6 +110,34 @@ def add_alias_arg_to_parser(parser: argparse.ArgumentParser) -> None:
...
+    parser.add_argument(
+        "--manifest-path",
+        dest="manifest_path",
+        action="store",
+        type=Path,
+        help=help_msg,
+    )
+
+
+def add_package_name_and_package_version_to_parser(
```
njgheorghita: Would you mind splitting this into two different util functions `add_package_name_to_parser()` and `add_package_version_to_parser()`? I can't think off the top of my head a use case where we'd need one and not the other, but it seems cleaner to separate them.

ethpm_cli/parser.py
```diff
@@ -117,41 +147,49 @@ def add_alias_arg_to_parser(parser: argparse.ArgumentParser) -> None:
...
+    package_version = args.package_version
+
+    if args.manifest_path:
+        (package_name, package_version, manifest_uri) = pin_local_manifest(
+            args.manifest_path
+        )
+        cli_logger.info(
+            f"Retrieving manifest info from local file @ {args.manifest_path} "
+        )
+
```
njgheorghita, line 163: On the python EF team - we try to avoid mutation wherever possible - https://github.com/ethereum/snake-charmers-tactical-manual/issues/27 - If you could update this to use an if/else and avoid mutating `package_name`, `package_version` and `manifest_uri` that'd be better

njgheorghita:
I made a minor update to web3 - which broke some tests here 😬 - so you'll have to rebase off master again to fix the failing tests in circle.

corydickson:
Running into issues for end-to-end testing the release cmd. Getting a bit of a cryptic error about it not being able to locate the `keyfile`. Gonna investigate further

njgheorghita:
@corydickson Hmm - it's probably that the cli is looking for a keyfile - but one hasn't been set. To sign any txs when you use `--keyfile-password` it needs a valid encrypted keyfile to be stored via `ethpm auth`. See [here](https://ethpm-cli.readthedocs.io/en/latest/quickstart.html#wallet-account-optional) and [here](https://ethpm-cli.readthedocs.io/en/latest/commands.html#ethpm-auth). I'm guessing you'll have to set up a keyfile for the default web3 account before you can release any packages. LMK if this sounds about right or not

corydickson:
I think it might have something to do with my environment: inside the `test/cli` there is a new file called `test_release.py` with the following snippet:

```python
import pexpect

from ethpm_cli._utils.ipfs import pin_local_manifest
from ethpm_cli.main import ENTRY_DESCRIPTION
from ethpm_cli.commands.auth import (
    get_authorized_private_key,
    get_keyfile_path,
    import_keyfile,
)
# from ethpm_cli.commands.release import release_package

def test_release_local_manifest(test_assets_dir, keyfile_auth, keyfile):
    # Release a package from a local manifest
    child_keyfile = pexpect.spawn(f"ethpm auth --keyfile-path {keyfile}", timeout=15)
    # Anytime you expect from child_keyfile here it fails with EOF

    local_manifest_path = test_assets_dir / "owned" / "1.0.0.json"
    (package_name, package_version, manifest_uri) = pin_local_manifest(local_manifest_path)
    _, _, password = keyfile_auth

    child = pexpect.spawn(f"ethpm release --manifest-path {local_manifest_path} --keyfile-password {password}", timeout=15)
    child.expect(ENTRY_DESCRIPTION)
    child.expect("\r\n")
    # EOF error here 
    child.expect(f"Retrieving manifest info from local file @ {local_manifest_path} ")
    child.expect(f"{package_name} v{package_version} @ {manifest_uri} ")
```

This looks like the keyfile path is malformed from the POSIX path given see the line `before (last 100 chars)`:
```
            exc.__cause__ = None # in Python 3.x we can use "raise exc from None"
>           raise exc
E           pexpect.exceptions.EOF: End Of File (EOF). Empty string style platform.
E           <pexpect.pty_spawn.spawn object at 0x10d93aa90>
E           command: /../../../ethpm-cli/venv/bin/ethpm
E           args: ['/../../.../ethpm-cli/venv/bin/ethpm', 'release', '--manifest-path', '/.../.../.../ethpm-cli/tests/core/assets/owned/1.0.0.json', '--keyfile-password', 'bpassword']
E           buffer (last 100 chars): b''
E           before (last 100 chars): b'ytest-of-cory.dickson/pytest-46/test_release_local_manifest0/xdg_ethpmcli_dir/_ethpm_keyfile.json.\r\n'
E           after: <class 'pexpect.exceptions.EOF'>
```

corydickson:
I'll try creating a new keyfile and writing it in the xdg test directory

corydickson:
Fixed had to import the keyfile in the test working through the rest now

corydickson:
@njgheorghita May I get a final pass on the PR as it stands? I'm having trouble with managing the web3 backends. 

Modified the snippet above with a new deployed registry like in the `pm` module, then added and activated it through the CLI. But even after sending the `authorized_address` some ETH (currently using the `keyfile` and `keyfile_auth` fixture) and calling `ethpm release --manifest-path ...`:
```
b'e["error"])\r\nValueError: {\'code\': -32000, \'message\': \'insufficient funds for gas * price + value\'}\r\n'
```

Assuming this is because the `config.w3` is not the same as the new `w3` fixture placed in `test/cli/conftest.py` and the chains are different. Tried sending a transaction through the `config` backend but it doesn't have access to `eth_coinbase` and the `defaultAccount` is an empty object.

You mentioned this wasn't high priority but if you have any suggestions on how to move forward I can continue working on it 🙏

njgheorghita:
@corydickson Hmm, hard to tell exactly what might be the problem, though your explanation seems plausible - if you want to push your code into a draft pr I'd be happy to take a closer look at it. 

In the meantime #88 should be a more exciting and definitely more useful challenge if you're up for it.

PR #82, opened 2020-02-08
Quickstart docs

njgheorghita:
## What was wrong?
quickstart docs to provide a simple jumping-in point for new users.

#### Cute Animal Picture
image

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -50,7 +50,7 @@ Test that your keyfile has been properly stored.
    ethpm auth
    > Keyfile stored for address: 0x123abc....................
 
-You can now use your keyfile's password as the flag ``--keyfile-password`` for any ethPM command, and it will be used to automatically sign any transactions.
+You can now use your keyfile's password with the flag ``--keyfile-password`` for any ethPM command, and it will be used to automatically sign any transactions.
 
 
 Activate
@@ -72,7 +72,7 @@ To instantly interact with any verified contract on Etherscan, use an `Etherscan
 
 .. code-block:: bash
 
-   ethpm activate etherscan://0x123c0ntractaddr3ss890:1
+   ethpm activate etherscan://0x123v3r1f13dc0ntractaddr3ss890:1
 
 
 Install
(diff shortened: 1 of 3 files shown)


njgheorghita:
@corydickson If you have some spare time to give this a quick look through for any errors / confusing lines of thought - that'd be awesome!

docs/quickstart.rst
```diff
@@ -0,0 +1,163 @@
...
+   ethpm auth --keyfile-path KEYFILE_PATH
+
+Test that your keyfile has been properly stored.
+
+.. code-block:: bash
+
+   ethpm auth
+   > Keyfile stored for address: 0x123abc....................
+
+You can now use your keyfile's password as the flag ``--keyfile-password`` for any ethPM command, and it will be used to automatically sign any transactions.
```
corydickson: Nit: change 'as the flag' -> 'with the flag' as it sounds like your password could be used as an argument by typing `--[my-password]`

docs/quickstart.rst
```diff
@@ -0,0 +1,163 @@
...
+
+If you have the Solidity compiler installed on your machine, the best way to get started is with the manifest wizard. The wizard expects a project directory with the following structure. 
+
+- project/
+  - contracts/
+    - xxx.sol
+    - yyy.sol
+
+Pass in a path to your project directory under the ``--project-dir`` flag. The wizard will attempt to compile these contracts using the available ``solc`` on your machine. Then it will prompt you for package details.
+
```
corydickson, line 113: Do you think it's worth documenting a minimum required solc version? This way a user knows that the package they want to release must use the same pragma as the `registry` contract if they are to deploy their own.

njgheorghita: Good call. Just to clarify though, the minimum solc required version would just be whatever version is required to compile the contracts that live in the project directory - not the pragma required for the registry contract. The registry contract is already compiled and deployed to the blockchain - to release a package on the registry all you need is the `package_name`, `version` and `manifest_uri`

pipermerriam (approved):
Nice!

PR #84, opened 2020-02-10
Separate auth command into more useful subparsers

njgheorghita:
## What was wrong?
`ethpm auth` required users to link an already generated encrypted keyfile. Now, the cli will handle that for you.

A couple new commands introduced...
`ethpm auth` - displays currently authenticated addr
`ethpm auth link --keyfile-path ./path/to/keyfile` - will link an existing keyfile
`ethpm auth unlink` - will unwrite whatever keyfile is stored in the xdg dir
`ethpm auth init` - will display a warning about private key handling, then prompt user for private key / password and store encrypted keyfile json in xdg directory

#### Cute Animal Picture
image

### DIFF
--- a/ethpm_cli/commands/auth.py
+++ b/ethpm_cli/commands/auth.py
@@ -1,23 +1,104 @@
+import json
 from pathlib import Path
-import tempfile
 from typing import Any, Dict
 
 import eth_keyfile
 from eth_typing import Address
-from eth_utils import to_bytes
+from eth_utils import add_0x_prefix, to_bytes
 
+from ethpm_cli._utils.filesystem import atomic_replace
+from ethpm_cli._utils.input import parse_bool_flag
+from ethpm_cli._utils.logger import cli_logger
 from ethpm_cli._utils.xdg import get_xdg_ethpmcli_root
 from ethpm_cli.constants import KEYFILE_PATH
 from ethpm_cli.exceptions import AuthorizationError
+
+PRIVATE_KEY_WARNING = (
+    "Please be careful when using your private key, it is a sensitive piece of information.\n",
+    "~ ~ ~ Not your keys, not your crypto. ~ ~ ~\n",
+    "ethPM doesn't save your private key, but it is best practice to re-use an already encrypted ",
+    "keyfile and link it with `ethpm auth link` rather than regenerating a new encrypted keyfile.",
+)
+
+
+def link_keyfile(keyfile_path: Path) -> None:
+    if valid_keyfile_exists():
+        xdg_keyfile = get_keyfile_path()
+        cli_logger.info(
+            f"Keyfile detected at {xdg_keyfile}. Please use `ethpm auth unlink` to delete this "
+            "keyfile before linking a new one."
+        )
+    else:
+        import_keyfile(keyfile_path)
+        address = get_authorized_address()
+        cli_logger.info(
+            f"Keyfile stored for address: {address}\n"
+            "It's now available for use when its password is passed in with the "
+            "`--keyfile-password` flag."
+        )
+
+
+def unlink_keyfile() -> None:
+    if not valid_keyfile_exists():
+        cli_logger.info("Unable to unlink keyfile: empty keyfile found.")
+    else:
+        keyfile_path = get_keyfile_path()
+        address = get_authorized_address()
+        keyfile_path.write_text("")
+        cli_logger.info(f"Keyfile removed for address: {address}")
+
+
+def init_keyfile() -> None:
+    if valid_keyfile_exists():
+        cli_logger.info(
+            f"Keyfile detected. Please use `ethpm auth unlink` to delete this "
+            "keyfile before initializing a new one."
+        )
+        return
+
+    cli_logger.info(PRIVATE_KEY_WARNING)
+    agreement = parse_bool_flag(
+        "Are you sure you want to proceed with initializing a keyfile? "
+    )
+    if not agreement:
+        cli_logger.info("Aborting keyfile initialization.")
+        return
+
+    private_key = to_bytes(text=input("Please enter your 32-length private key: "))
+    validate_private_key_length(private_key)
+    password = to_bytes(
+        text=input(
+            "Please enter a password to encrypt your keyfile with (Don't forget this password!): "
+        )
+    )
+    keyfile_json = eth_keyfile.create_keyfile_json(private_key, password)
+    ethpm_xdg_root = get_xdg_ethpmcli_root()
+    ethpm_cli_keyfile_path = ethpm_xdg_root / KEYFILE_PATH
+    with atomic_replace(ethpm_cli_keyfile_path) as file:
+        file.write(json.dumps(keyfile_json))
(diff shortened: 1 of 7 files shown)


ethpm_cli/commands/auth.py
```diff
@@ -1,23 +1,104 @@
...
+PRIVATE_KEY_WARNING = (
+    "Please be careful when using your private key, it is a sensitive piece of information.\n",
+    "~ ~ ~ Not your keys, not your crypto. ~ ~ ~\n",
+    "ethPM doesn't save your private key, but it is best practice to re-use an already encrypted ",
+    "keyfile and link it with `ethpm auth link` rather than regenerating a new encrypted keyfile.",
+)
+
+
+def link_keyfile(keyfile_path: Path) -> None:
+    if valid_keyfile_exists():
```
pipermerriam, line 25: I find it odd that this branch of the logic completely ignores the function argument `keyfile_path`.  Haven't gotten to the place where this function is used yet so maybe this is benign...

ethpm_cli/commands/auth.py
```diff
@@ -1,23 +1,104 @@
...
+        cli_logger.info(
+            f"Keyfile stored for address: {address}\n"
+            "It's now available for use when its password is passed in with the "
+            "`--keyfile-password` flag."
+        )
+
+
+def unlink_keyfile() -> None:
+    if not valid_keyfile_exists():
+        cli_logger.info("Unable to unlink keyfile: empty keyfile found.")
```
pipermerriam, line 43: The phrase: "empty keyfile found" might be misleading.  Should this say something like "no keyfile found" and maybe include the search path so that they know where it was supposed to be?

ethpm_cli/commands/auth.py
```diff
@@ -1,23 +1,104 @@
...
+    else:
+        keyfile_path = get_keyfile_path()
+        address = get_authorized_address()
+        keyfile_path.write_text("")
```
pipermerriam, line 47: I find this behavior surprising.  I would expect it to remove the file...

ethpm_cli/commands/auth.py
```diff
@@ -1,23 +1,104 @@
...
+        cli_logger.info(f"Keyfile removed for address: {address}")
+
+
+def init_keyfile() -> None:
+    if valid_keyfile_exists():
+        cli_logger.info(
+            f"Keyfile detected. Please use `ethpm auth unlink` to delete this "
+            "keyfile before initializing a new one."
+        )
+        return
```
pipermerriam, line 57: I think expected behavior would be for the process to return a non-zero return code so that if this is run in some sort of script that this logic branch produces a detectable failure.

ethpm_cli/commands/auth.py
```diff
@@ -1,23 +1,104 @@
...
+
+    cli_logger.info(PRIVATE_KEY_WARNING)
+    agreement = parse_bool_flag(
+        "Are you sure you want to proceed with initializing a keyfile? "
+    )
+    if not agreement:
+        cli_logger.info("Aborting keyfile initialization.")
+        return
```
pipermerriam, line 65: Same here, this should probably result in the CLI exiting with a non-zero return code..

ethpm_cli/commands/auth.py
```diff
@@ -1,23 +1,104 @@
...
+
+    private_key = to_bytes(text=input("Please enter your 32-length private key: "))
```
pipermerriam, line 67: I don't think this is ok.  I know you are probably expecting people to copy/paste which *might* be fine if it could be enforced, but this will also result in people typing the key in, or pasting it with an accidental newline at the end or typing it in its hexidecimal representation, or any number of insane things we can't think of....  I think that if someone wants to bring in a raw private key we should require them to do the heavy lifting of creating the keyfile themselves so that we're not responsible for mistakes by encouraging this pattern... thoughts?

cc @holiman

ethpm_cli/commands/auth.py
```diff
@@ -1,23 +1,104 @@
...
+    password = to_bytes(
+        text=input(
+            "Please enter a password to encrypt your keyfile with (Don't forget this password!): "
+        )
+    )
+    keyfile_json = eth_keyfile.create_keyfile_json(private_key, password)
+    ethpm_xdg_root = get_xdg_ethpmcli_root()
+    ethpm_cli_keyfile_path = ethpm_xdg_root / KEYFILE_PATH
+    with atomic_replace(ethpm_cli_keyfile_path) as file:
+        file.write(json.dumps(keyfile_json))
```
pipermerriam, line 78: *nitpick* you can change this to `json.dump(keyfile_json, file)` which is maybe just *slightly* more idiomatic.

ethpm_cli/commands/auth.py
```diff
@@ -1,23 +1,104 @@
...
+
 
 def import_keyfile(keyfile_path: Path) -> None:
     validate_keyfile(keyfile_path)
     ethpm_xdg_root = get_xdg_ethpmcli_root()
     ethpm_cli_keyfile_path = ethpm_xdg_root / KEYFILE_PATH
-    tmp_keyfile = Path(tempfile.NamedTemporaryFile().name)
-    tmp_keyfile.write_text(keyfile_path.read_text())
-    tmp_keyfile.replace(ethpm_cli_keyfile_path)
+    with atomic_replace(ethpm_cli_keyfile_path) as file:
```
pipermerriam, line 100: *orthogonal*

I'm thinking we should move this utility into `eth-utils` as `atomic_replace` is a pretty useful idiom.

ethpm_cli/commands/auth.py
```diff
@@ -49,7 +130,7 @@ def get_authorized_address() -> Address:
     Returns the address associated with stored keyfile. No password required.
     """
     keyfile = get_keyfile_data()
-    return keyfile["address"]
+    return add_0x_prefix(keyfile["address"])
```
pipermerriam, line 133: thoughts on making this a checksum address?  It's getting logged in a bunch of places which means it will end up in people's clipboards.  Since people will need to fund this address, they're likely to copy/pasta it to send funds...

ethpm_cli/parser.py
```diff
@@ -165,32 +170,71 @@ def auth_action(args: argparse.Namespace) -> None:
...
+        link_keyfile(args.keyfile_path)
+
+
+def auth_unlink_action(args: argparse.Namespace) -> None:
+    Config(args)
+    unlink_keyfile()
+
+
+def auth_init_action(args: argparse.Namespace) -> None:
+    Config(args)
```
pipermerriam, line 197: This is a code smell to me since I'm assuming you are relying on some sort of side-effect action that happens by instantiating this object?  If so I would advocate making this API more explicit so that it is clear what is happening here and why.  At minimum a code comment, but I think that the side effect nature of object instantiation isn't ideal and should potentially be moved into something like a `classmethod` which handles the side-effect stuff and keep the validation checks in the `__init__`...

pipermerriam (commented):
I'd like to give it another pass when ready.

PR #87, opened 2020-02-12
Add missing __init__ file

njgheorghita:
## What was wrong?
Sphinx-argparse (aka auto-doc cli commands) was broken in the docs on master but worked on local builds. Digging into it - it turns out there's no `__init__.py` file in `ethpm_cli/commands` - which also brought up some missed types.

#### Cute Animal Picture
image


njgheorghita:
merging since minor

Issue #88, opened 2020-02-17
ethpm registry explore [URI]


njgheorghita:
A new subcommand for `ethpm registry` that would accept any valid registry uri and display the available packages and their manifest uris that are available on the registry (similar display to `ethpm list`).

## How can it be fixed?
- A new command (`explore`) added to the `ethpm registry` command that requires a URI argument. 
- Then validate that this is a valid registry uri.
- Then make a call to the blockchain to fetch all available packages via the `web3.pm` module. 
- Display the available packages in a manner similar to `ethpm list`. 

For testing it would be nice to have a local chain spun up to test against rather than using infura and the mainnet - we should be able to avoid some of the challenges ran into [here](https://github.com/ethpm/ethpm-cli/pull/81#issuecomment-586631236) since we're just reading from the blockchain rather than cutting a new release.

corydickson:
It might also be useful to allow the user to provide an `alias` of a registry that's already added. Some immediate blockers might be related to #85 since we would not be able to guarantee a properly formatted URI. Where as it's easier to validate a registry_uri of the format `erc1319://0xDEADbeef`

I'll start on this initial approach and see how far I can get

njgheorghita:
Great! I imagine this would be fairly similar to `activate_registry()` in `commands/registry.py` - in terms of how we handle using an alias. Agreed that it might be funky until #85 drops - but if we follow the `activate_registry` pattern - then integrating these down the road shouldn't be too complex.... I hope

PR #92, opened 2020-02-24
Registry explore cmd

corydickson:
## What was wrong?

Issue #88 

## How was it fixed?

Added a new subparser to `registry` that accepts a valid URI and prints the total releases with their corresponding `version` and `manifest_uri` as well as the package name. 

#### Cute Animal Picture

### DIFF
--- a/ethpm_cli/commands/registry.py
+++ b/ethpm_cli/commands/registry.py
@@ -7,6 +7,7 @@ from eth_utils import to_tuple
 from eth_utils.toolz import assoc, assoc_in, dissoc
 from ethpm.backends.registry import is_valid_registry_uri, parse_registry_uri
 from ethpm.constants import SUPPORTED_CHAIN_IDS
+from web3 import Web3
 
 from ethpm_cli._utils.filesystem import atomic_replace
 from ethpm_cli._utils.logger import cli_logger
@@ -117,7 +118,7 @@ def explore_registry(uri_or_alias: str, config: Config) -> None:
         parsed_registry_uri = parse_registry_uri(registry.uri)
         config.w3.pm.set_registry(parsed_registry_uri.address)
     package_names = config.w3.pm.get_all_package_names()
-    display_packages(package_names, config)
+    display_packages(package_names, config.w3)
 
 
 def resolve_uri_or_alias(uri_or_alias: str, store_path: Path) -> StoredRegistry:
@@ -218,13 +219,12 @@ def generate_registry_store_data(
     }
 
 
-def display_packages(all_package_names: Iterable[str], config: Config) -> None:
+def display_packages(all_package_names: Iterable[str], w3: Web3) -> None:
+    cli_logger.info(f"Packages in the registry: {w3.pm.get_package_count()}\n")
     for package_name in all_package_names:
-        all_releases = config.w3.pm.get_all_package_releases(package_name)
         cli_logger.info(f"Retrieving all releases for {bold_blue(package_name)}: \n")
+        all_releases = w3.pm.get_all_package_releases(package_name)
         for version, manifest_uri in all_releases:
             cli_logger.info(f"{bold_green(version)} --- ({bold_white(manifest_uri)})")
-        cli_logger.info(
-            f"Total releases: {config.w3.pm.get_release_count(package_name)}\n"
-        )
-    cli_logger.info(f"Packages in the registry: {config.w3.pm.get_package_count()}\n")
+
+        cli_logger.info(f"Total releases: {len(all_releases)}\n")
--- a/ethpm_cli/commands/registry.py
+++ b/ethpm_cli/commands/registry.py
@@ -114,21 +114,15 @@ def activate_registry(uri_or_alias: str, config: Config) -> None:
 
 def explore_registry(uri: str, config: Config) -> None:
     if is_valid_registry_uri(URI(uri)):
-        (
-            address,
-            chain_id,
-            pkg_name,
-            pkg_version,
-            namespaced_asset,
-            ens,
-        ) = parse_registry_uri(uri)
-        config.w3.enable_unstable_package_management_api()
-        config.w3.pm.set_registry(address)
+        parsed_registry_uri = parse_registry_uri(uri)
+        config.w3.pm.set_registry(parsed_registry_uri.address)
         package_names = config.w3.pm.get_all_package_names()
         display_packages(package_names, config)
 
     else:
-        raise UriNotSupportedError("Cannot resolve both an alias and registry uri.")
+        raise UriNotSupportedError(
+            f"Invalid URI: {uri}, unable to explore. Registry URI definition can be found at https://docs.ethpm.com/uris#registry-uris."  # noqa: E501
+        )
 
 
 def resolve_uri_or_alias(uri_or_alias: str, store_path: Path) -> StoredRegistry:
@@ -232,13 +226,10 @@ def generate_registry_store_data(
 def display_packages(all_package_names: Iterable[str], config: Config) -> None:
     for package_name in all_package_names:
         all_releases = config.w3.pm.get_all_package_releases(package_name)
-        cli_logger.info(f"Retrieving releases for {bold_blue(package_name)}: \n")
-        for release in all_releases:
-            (version, manifest_uri) = release
+        cli_logger.info(f"Retrieving all releases for {bold_blue(package_name)}: \n")
+        for version, manifest_uri in all_releases:
             cli_logger.info(f"{bold_green(version)} --- ({bold_white(manifest_uri)})")
         cli_logger.info(
-            f"Total: {config.w3.pm.get_release_count(package_name)}\n"
+            f"Total releases: {config.w3.pm.get_release_count(package_name)}\n"
         )
-    cli_logger.info(
-        f"Packages in the registry: {config.w3.pm.get_package_count()}\n"
-    )
+    cli_logger.info(f"Packages in the registry: {config.w3.pm.get_package_count()}\n")
--- a/ethpm_cli/commands/registry.py
+++ b/ethpm_cli/commands/registry.py
@@ -13,12 +13,7 @@ from ethpm_cli._utils.logger import cli_logger
 from ethpm_cli._utils.shellart import bold_blue, bold_green, bold_white
 from ethpm_cli.config import Config
 from ethpm_cli.constants import REGISTRY_STORE
-from ethpm_cli.exceptions import (
-    AmbigiousFileSystem,
-    AuthorizationError,
-    InstallError,
-    UriNotSupportedError,
-)
+from ethpm_cli.exceptions import AmbigiousFileSystem, AuthorizationError, InstallError
 
 
 class StoredRegistry(NamedTuple):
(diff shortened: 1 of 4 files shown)


ethpm_cli/commands/registry.py
```diff
@@ -107,6 +112,25 @@ def activate_registry(uri_or_alias: str, config: Config) -> None:
...
+def explore_registry(uri: str, config: Config) -> None:
+    if is_valid_registry_uri(URI(uri)):
+        (
+            address,
+            chain_id,
+            pkg_name,
+            pkg_version,
+            namespaced_asset,
+            ens,
+        ) = parse_registry_uri(uri)
```
corydickson: We shouldn't need any of these other than the `address` but was unable to find a suitable function, I'll use underscore to remove these unused variables

tests/cli/test_registry_cli.py
```diff
@@ -67,3 +67,10 @@ def test_ethpm_registry_commands():
     child_four.expect(r"\(active\)")
     child_four.expect("erc1319://0xabc0000000000000000000000000000000000000:3")
     child_four.expect("yours")
+
+    # test registry explore
+    child_five = pexpect.spawn(
+        f"ethpm registry explore erc1319://0x16763EaE3709e47eE6140507Ff84A61c23B0098A:1"
+    )
+    child_five.expect(ENTRY_DESCRIPTION)
+    child_five.expect("\r\n")
```
corydickson, line 78: If you like the current formatting I'll add additional tests for an existing registry live on mainnet, hopefully something with fewer entries 😭

corydickson:
I'll push a fix for the linter if the formatting looks good on the output and add the corresponding tests!

tests/cli/test_registry_cli.py
corydickson: Actually after some thought that might be a bit flaky in the future if people add packages and the order is not preserved... maybe we should just add a new test case that releases a local package to a fresh registry

ethpm_cli/commands/registry.py
njgheorghita: [`parse_registry_uri(uri)`](https://github.com/ethereum/web3.py/blob/master/ethpm/backends/registry.py#L88) returns a `RegistryURI` type, which is a custom `NamedTuple` -  so you should be able to do something like
```python
parsed_registry_uri = parse_registry_uri(uri)
```
and then access properties on it like
```python
parsed_registry_uri.address
```

ethpm_cli/commands/registry.py
```diff
@@ -107,6 +112,25 @@ def activate_registry(uri_or_alias: str, config: Config) -> None:
...
+        config.w3.enable_unstable_package_management_api()
```
njgheorghita: I'm not sure you need this line here - was it raising an error without it? The `enable_unstable_package_management_api()` flag should have been handled in [`config.py`](https://github.com/ethpm/ethpm-cli/blob/master/ethpm_cli/config.py#L113)

ethpm_cli/parser.py
```diff
@@ -141,6 +142,12 @@ def add_package_version_to_parser(
     )
 
 
+def add_uri_to_parser(parser: argparse.ArgumentParser, help_msg: str,) -> None:
```
njgheorghita: You can remove the trailing comma after `str` here. Trailing commas are typically only useful when there is only one argument per line aka
```python
things =  (
    parser: argparse.ArgumentParser,
    help_msg: str,
)
```

ethpm_cli/parser.py
```diff
@@ -323,6 +336,14 @@ def registry_remove_cmd(args: argparse.Namespace) -> None:
 )
 registry_activate_parser.set_defaults(func=registry_activate_cmd)
 
+# ethpm registry explore
+registry_explore_parser = registry_subparsers.add_parser(
+    "explore", help="Explore a registry's list of released packages and manifest uris.",
+)
+add_uri_to_parser(
+    registry_explore_parser, "Registry URI for target registry.",
```
njgheorghita, line 347: Were we thinking of also supporting any aliased registries? I think it'd be more useful than requiring the entire uri everytime you want to explore a registry.  imo `ethpm registry explore ens` > `ethpm registry explore ethpm://ens.snakecharmers.eth/` for example. I think we do something similar for `ethpm registry activate`

tests/core/test_registry.py
```diff
@@ -141,3 +142,8 @@ def test_unable_to_activate_nonexistent_aliased_registry(config):
     add_registry(URI_1, "mine", config)
     with pytest.raises(InstallError):
         activate_registry("other", config)
+
+
+def test_explore_uri_invalid_raises_error(config):
```
njgheorghita: This looks like a great place to use [`@pytest.mark.parametrize`](https://docs.pytest.org/en/latest/parametrize.html) - we're big fans of  this decorator and tend to [use](https://docs.pytest.org/en/latest/parametrize.html) it a [lot](https://github.com/ethpm/ethpm-cli/blob/master/tests/core/test_install.py#L29). It lets you test a lot of varied assumptions without a ton of bloat. I'd load it up with any useful or weird edge cases you want to test.

ethpm_cli/commands/registry.py
```diff
@@ -107,6 +112,25 @@ def activate_registry(uri_or_alias: str, config: Config) -> None:
...
+        config.w3.pm.set_registry(address)
+        package_names = config.w3.pm.get_all_package_names()
+        display_packages(package_names, config)
+
+    else:
+        raise UriNotSupportedError("Cannot resolve both an alias and registry uri.")
```
njgheorghita: I'm not sure I understand this error message. It seems as though we currently aren't accepting an alias? Also, I tend to use f-strings a lot in my error messages - displaying the incorrect variable in the mesage can be extremely useful for users and yourself later on.
Something more like...
```python
else:
    raise UriNotSupportedError(f"Invalid URI: {uri}, unable to explore. Registry URI definition can be found at https://docs.ethpm.com/uris#registry-uris.")
```

ethpm_cli/commands/registry.py
```diff
@@ -203,3 +227,18 @@ def generate_registry_store_data(
...
         "active": activate,
     }
+
+
+def display_packages(all_package_names: Iterable[str], config: Config) -> None:
+    for package_name in all_package_names:
+        all_releases = config.w3.pm.get_all_package_releases(package_name)
+        cli_logger.info(f"Retrieving releases for {bold_blue(package_name)}: \n")
+        for release in all_releases:
+            (version, manifest_uri) = release
```
njgheorghita: I think you can do `for version, manifest_uri in all_releases:`

ethpm_cli/commands/registry.py
```diff
@@ -203,3 +227,18 @@ def generate_registry_store_data(
...
+            cli_logger.info(f"{bold_green(version)} --- ({bold_white(manifest_uri)})")
+        cli_logger.info(
+            f"Total: {config.w3.pm.get_release_count(package_name)}\n"
```
njgheorghita: nitpick be a bit more descriptive here - total what?

tests/cli/test_registry_cli.py
njgheorghita: Looks good to me! That would be ideal - but I think that would require us to have the tester chain up and running? But, we get a free out of jail card here b/c of the way the registry contract is written it returns the package ids in the same order every time - new ones are at the end of the list. So, you should be safe to use one of the mainnet registries  and just test that the first handful of packages are displayed.

njgheorghita (changes requested):
It's a great start, thanks! Just some changes to be made - let me know if you have any questions

ethpm_cli/parser.py
corydickson: Ahh okay I see the desired behavior; I'll remove the error thrown directly in `explore_registry` and just use `resolve_uri_or_alias`

ethpm_cli/commands/registry.py
corydickson: See the other comment below but I think it's okay to get rid of this if we are going to use existing error messages?

tests/core/test_registry.py
corydickson: To follow up the rest of the review would we want to even test that this function is throwing an error here anymore? Instead make it something like `test_resolve_uri_or_alias_raises_error` and try to hit all the cases

ethpm_cli/commands/registry.py
njgheorghita: Yep, that looks like it should be fine

tests/core/test_registry.py
njgheorghita: Yeah that's a better idea, i'd just write tests that `resolve_uri_or_alias` fails/passes for all the various edge cases.

corydickson:
Should have the tests ready this weekend, bit busy this week sorry for the delay

njgheorghita:
@corydickson No worries! just ping me whenever this is ready for a review

corydickson:
@njgheorghita Thank you for the patience, should be ready for review!

corydickson:
May I make the changes suggested by the linter here?

tests/cli/test_registry_cli.py
```diff
@@ -67,3 +68,42 @@ def test_ethpm_registry_commands():
...
+        f"Retrieving all releases for {bold_blue('augurreputation-rep')}: "
+    )
+    child_five.expect("\r\n\r\n")
+    child_five.expect_exact(
+        f"{bold_green('1.0.0')} --- ({bold_white('ipfs://QmXsunnsZWYRfCN1YyfJJoFLeRxZrSBhQC9b2HeXvYHrAH')})"  # noqa: 501
+    )
+    child_five.expect("\r\n")
+    child_five.expect(f"Total releases: 1")
+    child_five.expect("\r\n\r\n")
+    child_five.expect_exact(f"Retrieving all releases for {bold_blue('dai-dai')}: ")
```
njgheorghita, line 93: Oh cool! Wasn't aware of `expect_exact`

tests/core/test_registry.py
```diff
@@ -141,3 +142,50 @@ def test_unable_to_activate_nonexistent_aliased_registry(config):
...
+)
+def test_resolve_uri_or_alias_raises_error(config, uri, alias, valid, expected):
+    store_path = config.xdg_ethpmcli_root / REGISTRY_STORE
+    add_registry(uri, alias, config)
+    if valid:
+        registry = resolve_uri_or_alias(alias, store_path)
+        assert registry.alias == expected["alias"]
+        registry = resolve_uri_or_alias(uri, store_path)
+        assert registry.uri == expected["uri"]
+    else:
```
njgheorghita: I'm sure I've given you conflicting rules of thumbs on how to do this, but imo - `if` statements are a good indicator that a test should be split up - especially if it's for valid/invalid cases. In this instance I'd split this test b/w the valid/invalid cases.

ethpm_cli/commands/registry.py
```diff
@@ -203,3 +216,15 @@ def generate_registry_store_data(
         "alias": alias,
         "active": activate,
     }
+
+
+def display_packages(all_package_names: Iterable[str], config: Config) -> None:
```
njgheorghita: You could probably get away with just passing the `w3` that exists on `config` here rather than the entire `config` instance - just to tidy things up a bit.

ethpm_cli/commands/registry.py
```diff
@@ -203,3 +216,15 @@ def generate_registry_store_data(
...
+    for package_name in all_package_names:
+        all_releases = config.w3.pm.get_all_package_releases(package_name)
+        cli_logger.info(f"Retrieving all releases for {bold_blue(package_name)}: \n")
```
njgheorghita, line 225: I'd put line 224 above 223 - since the call to w3 to fetch all of the releases could take a couple of seconds, it would be better to have this message displayed before that call is made.

ethpm_cli/commands/registry.py
```diff
@@ -203,3 +216,15 @@ def generate_registry_store_data(
...
+        for version, manifest_uri in all_releases:
+            cli_logger.info(f"{bold_green(version)} --- ({bold_white(manifest_uri)})")
+        cli_logger.info(
+            f"Total releases: {config.w3.pm.get_release_count(package_name)}\n"
```
njgheorghita: You should be able to get away without this `w3` call and just use `len(all_releases)` to get the count.

ethpm_cli/commands/registry.py
```diff
@@ -203,3 +216,15 @@ def generate_registry_store_data(
...
+        )
+    cli_logger.info(f"Packages in the registry: {config.w3.pm.get_package_count()}\n")
```
njgheorghita: Any particular reason that this comes at the end of the function? It might be more useful at the beginning - like if there are 1000 packages available, it'd be better to know before all of the calls for each particular package rather than after all 1000 calls have been made and displayed.

njgheorghita (approved):
Looks good to me! Just a couple nitpicks left. Thanks for wrapping this one up - hope you're well!

njgheorghita:
Also the mypy fixes are in master now if you want to rebase against master to fix the linting errors.

tests/core/test_registry.py
corydickson: Got it, I'll split these up

ethpm_cli/commands/registry.py
corydickson: Good call

ethpm_cli/commands/registry.py
corydickson: No particular reason but your suggestion makes a lot more sense. I'm not going to add a test for this line though to prevent it from being flaky in the future

corydickson:
lol sorry for the fp was having issues... should be rebased correctly

njgheorghita:
Looks good, thanks @corydickson !!

PR #97, opened 2020-07-06
V3

njgheorghita:
## What was wrong?
Update CLI tool to support ethpm V3. 

Blocked until https://github.com/ethereum/web3.py/pull/1652 is available as a release.

#### Cute Animal Picture
image

PR #98, opened 2020-07-06
Update dai manifest for testing

njgheorghita:
## What was wrong?
In order to test `ethpm activate` with a github blob uri, the `dai` manifest needs updating.

## How was it fixed?
Updated `tests/core/assets/dai/_ethpm_packages/dai/manifest.json` from v2 to v3.

#### Cute Animal Picture
image


njgheorghita:
Merging since minor update and we need to get this asset merged to generate a blob uri to test against. This will break the tests on master, but only temporarily until #97 lands.

PR #99, opened 2020-08-13
Update explore registry to display registry owner

njgheorghita:
## What was wrong?
- When listing the available packages on a registry via `ethpm registry explore 0x123...`, it now displays the controlling address of the registry, to match the output in truffle.
- Also a minor bugfix

#### Cute Animal Picture
image


njgheorghita:
merging since minor addition / bugfix

